PR #4, opened 2023-10-18
Initial PR

AnotherDaniel:
Initial PR for Rust port of uprotocol-sdk


evshary:
Hello @AnotherDaniel

I'm curious about the API definition. I found that you create the API `register` for trait `UTransport`
https://github.com/SoftwareDefinedVehicle/uprotocol-sdk-rust_fork/blob/main/src/transport/datamodel/utransport.rs#L19

However, in the interface `UTransport` in Python, the API is `authenticate`
https://github.com/eclipse-uprotocol/uprotocol-java/blob/main/src/main/java/org/eclipse/uprotocol/transport/UTransport.java#L50C13-L50C25

Is there any reason?

By the way, for `register_listener` and `unregister_listener` in trait `UTransport`, it might be better to have the associated type instead of `dyn UListener`, since we can't use the unsized arguments directly in the function.

Something like this:

```rust
pub trait UTransport {
    type L: UListener;
....
    /// Register a method that will be called when a message comes in on the specific topic.
    fn register_listener(&self, topic: UUri, listener: Self::L) -> UStatus;

    /// Unregister a method on a topic. Messages arriving on this topic will no longer be processed by this listener.
    fn unregister_listener(&self, topic: UUri, listener: Self::L) -> UStatus;
}
```

AnotherDaniel:
Hi @evshary - thank you for looking into this.

Your first question has a simple answer: temporal divergence ;-)
The Java SDK made a large number of changes very recently (implementing what's essentially uProtocol spec version 1.5), while my Rust PR has been submitted before that recent batch of changes. I plan to update the Rust SDK to match the Java version asap - ideally after the PR has been merged (because that would make my life easer), but I'll start on converging this week.

Your second point - thank you for the input, will put that on my changelist!

stevenhartley:
@AnotherDaniel I'm merging your change to the 1.5 branch to make your development simpler and so you can continue to align the rust sdk to the java SDK. Then we need to figure out the container/build dependencies and I want to work out the protoc assets issues now (how to build and where to fetch the pre-built assets like we do with Java/Kotlin from maven central)

PR #7, opened 2023-11-28
Uprotocol 1.5

AnotherDaniel:
A large number of changes all across the codebase to get up to uProtocol v1.5 spec / alignment with v1.5 Java SDK.


tests/uris.json
stevenhartley, line 1: FYI this will move to uprotocol-tck when it is ready.

src/transport/datamodel/utransport.rs
```diff
@@ -11,19 +11,61 @@
...
+
+/// `UTransport` is the uP-L1 interface that provides a common API for uE developers to send and receive messages.
+///
+/// Implementations of `UTransport` contain the details for connecting to the underlying transport technology and
+/// sending `UMessage` using the configured technology. For more information, please refer to
+/// [uProtocol Specification](https://github.com/eclipse-uprotocol/uprotocol-spec/blob/main/up-l1/README.adoc).
+#[async_trait]
 pub trait UTransport {
-    /// API to register the calling uE with the underlying transport implementation.
-    fn register(&self, uentity: UEntity, token: &[u8]) -> UStatus;
```
stevenhartley, line 19: What is this register API?

src/transport/datamodel/utransport.rs
```diff
@@ -11,19 +11,61 @@
...
+    /// by this listener.
+    ///
+    /// # Arguments
+    /// * `topic` - Resolved `UUri` for where the listener was registered to receive messages from.
+    /// * `listener` - The method to execute to process the data for the topic.
+    ///
+    /// # Returns
+    /// Returns `OKSTATUS` if the listener is unregistered correctly, otherwise returns `FAILSTATUS`
+    /// with the appropriate failure.
+    async fn unregister_listener(&self, topic: UUri, listener: Sender<UMessage>) -> UStatus;
```
stevenhartley, line 70: it doesn't have to be async

src/types/validationresult.rs
```diff
@@ -58,12 +76,17 @@ impl ValidationResult {
         ValidationResult::Failure(message.to_string())
     }
 
-    pub fn to_status(&self) -> UStatus {
+    pub fn to_status(&self) -> Status {
```
stevenhartley, line 79: You should use UStatus and not Status, I want to get rid of the googleapis dependency please

stevenhartley (changes requested):
Please see comments. Hard to look at each line but 99% of it looks great!

src/transport/datamodel/utransport.rs
AnotherDaniel: That was what was in the Java SDK months ago... but gone now, replaced by authenticate

.vscode/settings.json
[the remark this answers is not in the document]
AnotherDaniel: Done

src/transport/datamodel/utransport.rs
```diff
@@ -11,19 +11,61 @@
...
+    ///
+    /// # Arguments
+    /// * `topic` - Resolved `UUri` topic to send the payload to.
+    /// * `payload` - Actual payload.
+    /// * `attributes` - Additional transport attributes.
+    ///
+    /// # Returns
+    /// Returns `OKSTATUS` if the payload has been successfully sent (ACK'ed), otherwise returns
+    /// `FAILSTATUS` with the appropriate failure.
+    async fn send(&self, topic: UUri, payload: UPayload, attributes: UAttributes) -> UStatus;
```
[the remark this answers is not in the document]
AnotherDaniel: The reason why I didn't pick this up yet is that there exists a UStatus::Ok - so what would it mean if I get an Err() from a function that then contains an ::Ok value? ...

README.md
```diff
@@ -20,7 +20,7 @@ __Note:__ the SDK currently builds two protobuf messages from upstream cloudeven
 - `cloudevents.proto` from the [CNCF CloudEvents specification project](https://github.com/cloudevents/spec/blob/main/cloudevents/formats/cloudevents.proto) and
 - `status.proto` from [Google APIs](https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto).
```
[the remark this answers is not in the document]
AnotherDaniel: Done

build.rs
```diff
@@ -20,8 +21,23 @@ fn main() -> std::io::Result<()> {
...
+            "https://raw.githubusercontent.com/googleapis/googleapis/master/google/rpc/status.proto",
+            "https://raw.githubusercontent.com/googleapis/googleapis/master/google/rpc/error_details.proto",
+
+            // cloudevent proto definitions
+            "https://raw.githubusercontent.com/cloudevents/spec/main/cloudevents/formats/cloudevents.proto", 
+
+            // uProtocol-project proto definitions
+            "https://raw.githubusercontent.com/eclipse-uprotocol/uprotocol-core-api/main/src/main/proto/uuid.proto",
+            "https://raw.githubusercontent.com/eclipse-uprotocol/uprotocol-core-api/main/src/main/proto/uri.proto",
+            "https://raw.githubusercontent.com/eclipse-uprotocol/uprotocol-core-api/main/src/main/proto/uattributes.proto",
```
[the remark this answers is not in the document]
AnotherDaniel: The build.rs script calls protoc to compile this into corresponding Rust stubs during build; all of these transient artifacts are located in the target/ build directories.

src/types/validationresult.rs
AnotherDaniel: Done - I actually missed this file during the migration.

src/transport/datamodel/utransport.rs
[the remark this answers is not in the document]
AnotherDaniel: But UStatus contains a UCode, which can be OK. 
I'm not fundamentally against doing Rust-idiomatic error handling - but I'm not happy about the potential semantics of an Error result that can contain an OK code...

src/cloudevent/datamodel/ucloudeventattributes.rs
```diff
@@ -69,7 +39,7 @@ impl UCloudEventAttributes {
         &self.hash
     }
 
-    pub fn priority(&self) -> &Option<Priority> {
+    pub fn priority(&self) -> &Option<UPriority> {
```
Mallets, line 42: In general it's better to return `Option<&T>` instead of `&Option<T>`.
If the function returns `Option<&T>`, then `T` has the freedom to live either on the heap or on the stack (like in this case). I.e., `T` or `Box<T>` can be both referenced with `&T`.
Instead, if the function returns `&Option<T>`, then `T` needs to live on the stack. If you want to move it to the heap (e.g. box-it), you will need to change the function signature to `&Option<Box<T>>`.

src/proto/uprotocol/uauthority.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    pub fn has_ip(authority: &UAuthority) -> bool {
+        matches!(authority.remote, Some(Remote::Ip(_)))
+    }
+
+    pub fn has_id(authority: &UAuthority) -> bool {
+        matches!(authority.remote, Some(Remote::Id(_)))
+    }
+
+    pub fn get_name(authority: &UAuthority) -> Option<&String> {
```
Mallets, line 30: When returning a reference to a `String` it's usually better to return `&str`.

src/proto/uprotocol/uauthority.rs
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    pub fn get_ip(authority: &UAuthority) -> Option<&Vec<u8>> {
+        match &authority.remote {
+            Some(Remote::Ip(ip)) => Some(ip),
+            _ => None,
+        }
+    }
+
+    pub fn get_id(authority: &UAuthority) -> Option<&Vec<u8>> {
```
Mallets, line 44: It's usually preferable to return a slice `&[T]` instead of a reference to a vector `&Vec<T>`.

src/proto/uprotocol/uauthority.rs
```diff
@@ -0,0 +1,65 @@
...
+            _ => None,
+        }
+    }
+
+    pub fn set_name(authority: &mut UAuthority, name: String) -> &mut UAuthority {
+        authority.remote = Some(Remote::Name(name));
+        authority
+    }
+
+    pub fn set_ip(authority: &mut UAuthority, ip: Vec<u8>) -> &mut UAuthority {
```
Mallets, line 56: Rust's standard library defines [IpAddr](https://doc.rust-lang.org/std/net/enum.IpAddr.html) data type, with comes along with all the data safety guarantees and associated functions for both `IPv4` and `IPv6`. 
At the time of writing, I don't have any additional insight on why `authority.remote` holds a `Vec<u8>` instead of `IpAddr`. IMHO it would be preferable to use the proper data type.

src/proto/uprotocol/uauthority.rs
```diff
@@ -0,0 +1,65 @@
...
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+use crate::uprotocol::{Remote, UAuthority};
+
+/// Helper functions to deal with UAuthority::Remote structure
+impl UAuthority {
+    pub fn has_name(authority: &UAuthority) -> bool {
```
Mallets, line 18: It seems that all functions defined in `impl UAuthority` take as first parameter either `&UAuthority` or `&mut UAuthority`.
Why not taking `&self` and `&mut self` instead as first parameter?

src/proto/uprotocol/uauthority.rs
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    pub fn get_id(authority: &UAuthority) -> Option<&Vec<u8>> {
+        match &authority.remote {
+            Some(Remote::Id(id)) => Some(id),
+            _ => None,
+        }
+    }
+
+    pub fn set_name(authority: &mut UAuthority, name: String) -> &mut UAuthority {
```
Mallets, line 51: Changing the function to something like:
```rust
pub fn set_name<T>(authority: &mut UAuthority, name: T) -> &mut Self 
where T: Into<String> {
authority.remote = Some(Remote::Name(name.into()));
authority
}
```
would increase ergonomics. E.g.:
```rust
authrotity.set_name("my_name");
authrotity.set_name(String::from("my_name"));
```
are possible, while with the current definition only the `String::from` version is possible.

src/proto/uprotocol/uauthority.rs
AnotherDaniel: Agree - I will look into prost compile options whether there is a way to influence the generated code for this, but earliest tonight.

src/proto/uprotocol/upayload.rs
```diff
@@ -0,0 +1,68 @@
...
+use prost::Message;
+use prost_types::Any;
+use std::slice;
+
+use crate::uprotocol::{Data, UPayload};
+
+impl From<Any> for UPayload {
+    fn from(value: Any) -> Self {
+        let vec = value.encode_to_vec();
+        let len = vec.len() as i32;
```
Mallets, line 23: In Rust length is usually represented with an `usize`. Any reason on using an `i32`?

src/proto/uprotocol/umessagetype.rs
```diff
@@ -0,0 +1,44 @@
...
+
+use crate::uprotocol::UMessageType;
+
+impl From<UMessageType> for String {
+    fn from(value: UMessageType) -> Self {
+        value.to_string()
+    }
+}
+
+impl From<String> for UMessageType {
```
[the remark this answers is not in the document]
AnotherDaniel: Done

src/proto/uprotocol/uauthority.rs
AnotherDaniel: Done

src/proto/uprotocol/uauthority.rs
AnotherDaniel: Done

src/proto/uprotocol/uauthority.rs
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    pub fn get_name(authority: &UAuthority) -> Option<&String> {
+        match &authority.remote {
+            Some(Remote::Name(name)) => Some(name),
+            _ => None,
+        }
+    }
+
+    pub fn get_ip(authority: &UAuthority) -> Option<&Vec<u8>> {
```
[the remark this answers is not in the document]
AnotherDaniel: Done

src/proto/uprotocol/uauthority.rs
AnotherDaniel: Done

src/cloudevent/datamodel/ucloudeventattributes.rs
AnotherDaniel: Done

src/proto/uprotocol/upayload.rs
AnotherDaniel: The protobuf-generated UPayload type takes i32 for it's length field, that is the reason for the cast.

src/proto/uprotocol/uresource.rs
```diff
@@ -0,0 +1,57 @@
...
+
+use crate::uprotocol::UResource;
+
+impl UResource {
+    pub fn has_id(resource: &UResource) -> bool {
+        resource.id.is_some()
+    }
+
+    pub fn get_message(resource: &UResource) -> Option<&String> {
+        match &resource.message {
```
Mallets, line 22: The function should return `Option<&str>`.

To be more *rusty*, the whole match can be replaced with a [deref](https://doc.rust-lang.org/std/option/enum.Option.html#method.as_deref), e.g.:
```rust
resource.message.as_deref()
```

src/proto/uprotocol/uresource.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    pub fn get_instance(resource: &UResource) -> Option<&String> {
+        match &resource.instance {
+            Some(instance) => Some(instance),
+            _ => None,
+        }
+    }
+}
+
+impl From<String> for UResource {
```
Mallets: A lot of allocations happen here. 
I suggest to implement the `From<&str>` and not to collect the split into a `Vec<_>`.
Using iterators you can achieve the same result with the difference of allocating just the fields necessary to construct `UResource` (and not any intermediary step).

src/proto/uprotocol/upayload.rs
Mallets: Ok, then to be safe in terms of hidden bugs cause by casting, I would explicitly perform a conversion between `usize` and `i32` with a [TryFrom](https://doc.rust-lang.org/std/primitive.i32.html#impl-TryFrom%3Cusize%3E-for-i32). That would be able to catch the case of the value length to be greater of `i32::MAX`.

This leads also to the fact that the `From<Any>` trait implentation should be a `TryFrom<Any>` since it could fail?

src/proto/uprotocol/upayload.rs
```diff
@@ -0,0 +1,68 @@
...
+            Data::Value(bytes) => {
+                return Some(bytes.as_slice());
+            }
+        }
+    }
+    None
+}
+
+// Please no one use this...
+fn read_memory(address: u64, length: i32) -> &'static [u8] {
```
Mallets, line 60: Since correctness of this function needs to be ensured the `caller`, I would mark the whole function as `unsafe fn read_memory(...) -> ...`

src/rpc/rpcmapper.rs
```diff
@@ -75,26 +67,35 @@ impl RpcMapper {
...
+    where
+        T: prost::Message + Default,
+        F: Future<Output = RpcClientResult>,
+    {
+        match response_future.await {
+            Ok(payload) => {
+                // we got some useable response payload
+                let any: Any = payload.into();
+                if any != Any::default() {
+                    match T::decode(any.value.as_slice()) {
```
Mallets, line 80: This match could be replaced with a [map_err](https://doc.rust-lang.org/std/result/enum.Result.html#method.map_err).

src/rpc/rpcmapper.rs
```diff
@@ -348,107 +344,115 @@ mod tests {
...
-
-            Box::pin(std::future::ready(Ok(payload)))
+        async fn invoke_method(
+            _topic: UUri,
+            _payload: UPayload,
+            _attributes: UAttributes,
+        ) -> RpcClientResult {
+            let status = UStatus::fail_with_code(UCode::InvalidArgument, "boom");
+
+            let any = RpcMapper::pack_any(status).unwrap();
```
Mallets, line 377: Since the function returns a `RpcClientResult`, what about removing the `unwrap()` that could cause panic and bubble up the error?

src/rpc/rpcmapper.rs
```diff
@@ -348,107 +344,115 @@ mod tests {
...
-
-            Box::pin(std::future::ready(Ok(payload)))
+        async fn invoke_method(
+            _topic: UUri,
+            _payload: UPayload,
+            _attributes: UAttributes,
+        ) -> RpcClientResult {
+            let status = UStatus::fail_with_code(UCode::Ok, "all good");
+
+            let any = RpcMapper::pack_any(status).unwrap();
```
Mallets, line 395: Since the function returns a RpcClientResult, what about removing the unwrap() that could cause panic and bubble up the error?

src/transport/builder/uattributesbuilder.rs
```diff
@@ -0,0 +1,212 @@
...
+
+    /// Adds the time to live in milliseconds.
+    ///
+    /// # Arguments
+    /// * `ttl` - The time to live in milliseconds.
+    ///
+    /// # Returns
+    /// Returns the `UAttributesBuilder` instance with the configured ttl.
+    pub fn with_ttl(&mut self, ttl: u32) -> &mut UAttributesBuilder {
+        self.ttl = Some(ttl as i32);
```
Mallets, line 128: Here we could have an overflow if `ttl` is larger than `i32::MAX`. Should panic? Should return an error?

src/transport/datamodel/utransport.rs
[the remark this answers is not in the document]
Mallets: A `Result` is an `enum` under the hood, the main difference is that returning a `Result` would allow you to use the `?` to early return in case of errors + all the mapping functionalities which would you need to care about your self. If we want to go for a `Result`, and we want to differentiate different codes for `Ok` and `Err`, the function should then return `Result<UStatusOk, UStatusErr>`. If, there is only one possible case for `Ok`, then `Result<(), UStatus>` seems appropriate.

src/proto/uprotocol/upayload.rs
AnotherDaniel: TryFrom is not possible unfortunately, as that is already implemented as part of the prost-generated code. But I will adopt your point about doing an error-aware cast.

src/transport/datamodel/utransport.rs
Mallets: I would leave it as `async` in the same way as `register_listener` is. 
What I would be careful about is that `Sender<UMessage>` comes from `std::sync::mpsc::Sender`, which is `sync`. That means that the `send` operation could block the thread. Therefore, I believe it's wiser to use an `async sender`. E.g.: [flume](https://docs.rs/flume/latest/flume/) provides both a `sync` and `async` channel.

src/transport/validator/uattributesvalidator.rs
```diff
@@ -25,169 +26,177 @@ use std::time::SystemTime;
...
+            Some(t) if t > 0 => t,
+            Some(_) => return ValidationResult::Success,
+            None => 0,
+        };
+
+        if let Some(uuid) = &attributes.id {
+            if let Some(time) = UuidUtils::get_time(uuid) {
+                let delta = SystemTime::now()
+                    .duration_since(SystemTime::UNIX_EPOCH)
+                    .unwrap()
```
Mallets, line 81: The `unwrap()` could be replace with an early return into `ValidationResult`. In that way a panic can be avoided (unless a panic is the desired behaviour`).

src/transport/validator/uattributesvalidator.rs
```diff
@@ -25,169 +26,177 @@ use std::time::SystemTime;
...
     ///
-    /// Returns a `UStatus` indicating whether the URI is valid or not.
-    fn validate_sink(&self, attributes: &UAttributes) -> UStatus {
-        if let Some(sink) = &attributes.sink {
-            return UriValidator::validate(sink);
+    /// Returns a `ValidationResult` that is success or failed with a failure message.
+    fn validate_commstatus(&self, attributes: &UAttributes) -> ValidationResult {
+        if let Some(cs) = attributes.commstatus {
+            if UCode::try_from(cs).is_err() {
+                return ValidationResult::Failure("Invalid Communication Status Code".into());
```
Mallets, line 165: Would it be possible to include the `unknown` status code as part of the message error?

src/transport/validator/uattributesvalidator.rs
```diff
@@ -310,455 +339,513 @@ pub struct ResponseValidator;
...
     /// # Returns
     ///
-    /// Returns a `UStatus` indicating whether the `MessageType` is valid or not.
-    fn validate_type(&self, attributes: &UAttributes) -> UStatus {
-        if attributes.message_type == UMessageType::Response {
-            UStatus::ok()
-        } else {
-            UStatus::fail_with_msg_and_reason("Wrong Attribute Type", UCode::InvalidArgument)
+    /// Returns a `ValidationResult` that is success or failed with a failure message.
+    fn validate_type(&self, attributes: &UAttributes) -> ValidationResult {
```
Mallets, line 354: Based on the actual implementation of the function, I believe the `ValidationResult` can be simply represented as a `Result<(), String>`. In general, I think it's worth considering using a crate like [anyhow](https://docs.rs/anyhow/latest/anyhow/index.html) in the SDK. By doing so, it is easier to ensure that multiple results can be bubbled up with just `?`, significantly reducing the amount of boilerplate for error handling or conversion.

src/uri/validator/urivalidator.rs
```diff
@@ -77,163 +72,640 @@ impl UriValidator {
     /// # Returns
     ///
     /// Returns a `UStatus` containing either a success or a failure, along with the corresponding error message.
-    pub fn validate_rpc_response(uri: &UUri) -> UStatus {
+    pub fn validate_rpc_response(uri: &UUri) -> ValidationResult {
         let status = Self::validate(uri);
```
Mallets, line 76: For example, if `validate` would return a `Result<(), ValidationErr>` (or anyhow error), this line could simply become:
```rust
Self::validate(uri)?;
```
and the next
```
if status.is_failure() {
  return status;
}
```
could be removed.

src/uuid/builder/uuidbuilder.rs
```diff
@@ -128,42 +123,40 @@ impl UUIDFactory for UUIDv6Factory {
 /// | var        | MUST be the The 2 bit variant defined by Section 4.1 of RFC
 /// | rand_b     | MUST 62 bits random number that is generated at initialization time of the uE only and reused otherwise |
 
-pub struct UUIDv8Factory {
+pub struct UUIDv8Builder {
     msb: Arc<Mutex<u64>>,
```
Mallets, line 127: Could an `AtomicU64` be used instead of an `Arc<Mutext<u64>>`?

src/uuid/builder/uuidutils.rs
```diff
@@ -31,8 +33,8 @@ impl UuidUtils {
     /// # Returns
     ///
     /// * `Vec<u8>` - The byte array representation of the UUID
-    pub fn to_bytes(uuid: &Uuid) -> Vec<u8> {
-        uuid.as_bytes().to_vec()
+    pub fn to_bytes(uuid: &uproto_Uuid) -> Vec<u8> {
+        Uuid::from(uuid.clone()).as_bytes().to_vec()
```
Mallets, line 37: I'm not 100% sure, but I seems there is some extra copy here that could be avoided.

src/uuid/builder/uuidutils.rs
```diff
@@ -44,8 +46,8 @@ impl UuidUtils {
     /// # Returns
     ///
     /// * `Result<Uuid, Error>` - UUID object built from the byte array
-    pub fn from_bytes(bytes: &[u8; 16]) -> Result<Uuid, Error> {
-        Ok(Uuid::from_bytes(*bytes))
+    pub fn from_bytes(bytes: &[u8; 16]) -> Result<uproto_Uuid, Error> {
```
Mallets, line 49: In Rust, usually this kind of operations are implemented via the `TryFrom` trait.

src/uuid/builder/uuidutils.rs
```diff
@@ -57,8 +59,11 @@ impl UuidUtils {
     /// # Returns
     ///
     /// * `Result<Uuid, Error>` - The UUID object representation of the string
-    pub fn from_string(uuid_str: &str) -> Result<Uuid, Error> {
-        Uuid::from_str(uuid_str)
+    pub fn from_string(uuid_str: &str) -> Result<uproto_Uuid, Error> {
```
Mallets, line 62: Minor comment about Rust naming convention.
The word *string* usually refers to `String` type, while `str` is used to refer to `&str` type.

src/uuid/builder/uuidutils.rs
```diff
@@ -156,3 +165,34 @@ impl UuidUtils {
         }
     }
 }
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Version {
+    /// An unknown version.
+    VersionUnknown = 0,
```
Mallets, line 172: Minor comment about naming convention. 
The variant name usually does not repeat the enum name.
E.g.: `Version::Unknown` instead of `Version::VersionUnknown`.

src/uuid/builder/uuidutils.rs
```diff
@@ -156,3 +165,34 @@ impl UuidUtils {
...
+        match value {
+            0 => Some(Version::VersionUnknown),
+            4 => Some(Version::VersionRandomBased),
+            6 => Some(Version::VersionTimeOrdered),
+            8 => Some(Version::VersionUprotocol),
+            _ => None,
+        }
+    }
+
+    /// Returns the integer representation of the version.
```
Mallets, line 194: You can ensure that enum variants are represented as `usize` by prepending `#[repr(usize)]` to enum definition ([see doc](https://doc.rust-lang.org/nomicon/other-reprs.html#repru-repri)).
If you do so, everyone can safely cast the enum to usize.

src/uuid/serializer/longuuidserializer.rs
```diff
@@ -0,0 +1,34 @@
...
+ ********************************************************************************/
+
+use std::str::FromStr;
+use uuid::Uuid;
+
+use crate::uprotocol::Uuid as uproto_Uuid;
+use crate::uuid::serializer::uuidserializer::UuidSerializer;
+
+/// UUID Serializer interface used to serialize/deserialize UUIDs to/from a string
+pub struct LongUuidSerializer;
```
Mallets, line 21: The common Rust trait to implement string parsing is the [FromStr trait](https://doc.rust-lang.org/std/str/trait.FromStr.html).

src/transport/datamodel/utransport.rs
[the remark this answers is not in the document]
AnotherDaniel: done. But don't ask me what an Err(Ok) would mean... ;-)

src/proto/uprotocol/upayload.rs
AnotherDaniel: done, although this entire thing will have to change anyways I think. I shudder just looking at that code. Plus the unsafe{} now moved to another helper function one above this one, not optimal yet...

src/transport/builder/uattributesbuilder.rs
AnotherDaniel: yeah, the joys of vague type definitions... I made an attempt to improve this.

src/uuid/serializer/longuuidserializer.rs
AnotherDaniel: done, in proto.uprotocol.uuid.rs

src/uuid/builder/uuidutils.rs
AnotherDaniel: done

src/uuid/builder/uuidutils.rs
AnotherDaniel: done

src/uuid/builder/uuidutils.rs
AnotherDaniel: removed, via above comment

src/uuid/builder/uuidutils.rs
AnotherDaniel: done

src/uuid/builder/uuidutils.rs
AnotherDaniel: done

src/uuid/builder/uuidbuilder.rs
AnotherDaniel: done

src/uri/validator/urivalidator.rs
AnotherDaniel: this, I will tackle separately

src/transport/validator/uattributesvalidator.rs
AnotherDaniel: this, I will tackle separately

src/transport/validator/uattributesvalidator.rs
AnotherDaniel: done

src/transport/validator/uattributesvalidator.rs
AnotherDaniel: done

src/rpc/rpcmapper.rs
AnotherDaniel: done

src/rpc/rpcmapper.rs
AnotherDaniel: done

src/rpc/rpcmapper.rs
```diff
@@ -75,26 +67,35 @@ impl RpcMapper {
...
-                            Err(error) => Err(RpcMapperError::InvalidPayload(error.to_string())), // ... or not
-                        },
-                        // ... or not
-                        Err(error) => Err(RpcMapperError::ProtobufError(error.to_string())),
+    pub async fn map_response<T, F>(response_future: F) -> Result<T, RpcMapperError>
+    where
+        T: prost::Message + Default,
+        F: Future<Output = RpcClientResult>,
+    {
+        match response_future.await {
```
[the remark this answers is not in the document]
AnotherDaniel: done

src/rpc/rpcmapper.rs
AnotherDaniel: done

src/proto/uprotocol/uresource.rs
AnotherDaniel: done

src/uri/validator/urivalidator.rs
AnotherDaniel: done - I like the results, even though I might have missed some simplification opportunities

src/transport/datamodel/utransport.rs
```diff
@@ -11,19 +11,66 @@
...
+    /// * `topic` - Resolved `UUri` for where the message arrived via the underlying transport technology.
+    /// * `listener` - The method to execute to process the data for the topic.
+    ///
+    /// # Returns
+    /// Returns a Uuid on success that can be used for unregistering later, otherwise an Err(UStatus) with the appropriate failure information.
+    async fn register_listener(
+        &self,
+        topic: UUri,
+        listener: Sender<UMessage>,
+    ) -> Result<Uuid, UStatus>;
```
sophokles73, line 64: FMPOV the returned Ok value should not be required to be a Uuid. The underlying implementation should be able to determine if a UUID is indeed required to keep it unique within its scope. For the MQTT based ulink, a simple MQTT 5 Subscription Identifier will probably be sufficient.
I would therefore suggest to change the type from `Uuid` to `String`. An implementation can then still return a (serialized) UUID if it so wishes.

@Mallets I have experimented with the following while playing around with an MQTT5 based implementation:
```rust
async fn register_listener<T>(&self, topic: UUri, listener: T) -> Result<String, UStatus>
    where
        T: Fn(UMessage) + Send + 'static;
```
WDYT?

src/rpc/rpcmapper.rs
```diff
@@ -75,26 +67,22 @@ impl RpcMapper {
...
-                            Err(error) => Err(RpcMapperError::InvalidPayload(error.to_string())), // ... or not
-                        },
-                        // ... or not
-                        Err(error) => Err(RpcMapperError::ProtobufError(error.to_string())),
-                    }
-                }
-                Err(error) => Err(error),
-            }
-        })
+    pub async fn map_response<T, F>(response_future: F) -> Result<T, RpcMapperError>
```
sophokles73, line 70: I wonder if this function really needs to take a Future as param? Mapping the result seems to be straight forward and should not require an async method, should it? IMHO client code will invoke the service, await the response and then simply invoke this mapping function with the result (not the future) ...

AnotherDaniel: done

src/transport/datamodel/utransport.rs
AnotherDaniel: done

src/cloudevent/builder/ucloudeventutils.rs
```diff
@@ -144,7 +144,9 @@ impl UCloudEventUtils {
     /// otherwise a `None` is returned.
     pub fn get_ttl(event: &Event) -> Option<u32> {
         if let Some(ExtensionValue::Integer(ttl)) = event.extension("ttl") {
-            return Some(*ttl as u32);
+            if let Ok(ttl) = u32::try_from(*ttl) {
```
Mallets, line 147: Converting a `Result<T, E>` into an `Option<T>` can be easily done with the `ok()` function. See [here](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok).

src/cloudevent/serializer/cloudeventserializer.rs
```diff
@@ -11,12 +11,43 @@
...
+    ///
+    /// # Returns
+    ///
+    /// Returns `Result<Vec<u8>, SerializationError>` where `Ok(Vec<u8>)` is the serialized byte vector representation of the `CloudEvent`.
+    ///
+    /// # Errors
+    ///
+    /// Returns `Err(SerializationError)` if the serialization process fails. This may occur due to issues like
+    /// invalid event data, failure in encoding the event into bytes, or other serialization-specific errors.
+    fn serialize(&self, cloud_event: &CloudEvent) -> Result<Vec<u8>, SerializationError>;
```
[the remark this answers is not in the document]
Mallets: Actually, the problem here is that we are mixing the memory support buffer with the actual codec. As an example, the eclipse-zenoh project separates those into [buffers](https://github.com/eclipse-zenoh/zenoh/tree/master/commons/zenoh-buffers) and [codec](https://github.com/eclipse-zenoh/zenoh/tree/master/commons/zenoh-codec). That provides additional flexibility and control on memory allocation. Would a similar approach work as well for `uprotocol`?

src/cloudevent/builder/ucloudeventutils.rs
AnotherDaniel: done

src/cloudevent/serializer/cloudeventserializer.rs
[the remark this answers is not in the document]
Mallets: I believe we need to have a deeper chat and session... extending the same approach to other languages may be tricky since they can't leverage the power of Rust's traits.

src/transport/datamodel/utransport.rs
AnotherDaniel: done

build.rs
```diff
@@ -20,10 +21,20 @@ fn main() -> std::io::Result<()> {
...
 
     if let Err(err) = get_and_build_protos(
-        &["https://raw.githubusercontent.com/cloudevents/spec/main/cloudevents/formats/cloudevents.proto", 
-        "https://raw.githubusercontent.com/googleapis/googleapis/master/google/rpc/status.proto"]
+        &[
+            // cloudevent proto definitions
+            "https://raw.githubusercontent.com/cloudevents/spec/main/cloudevents/formats/cloudevents.proto", 
+
+            // uProtocol-project proto definitions
+            "https://raw.githubusercontent.com/eclipse-uprotocol/uprotocol-core-api/main/src/main/proto/uuid.proto",
```
sophokles73, line 29: in order to prevent these from becoming a *moving target* I would propose to pin a particular version (1.5.3?) instead of using whatever is currently on `main` ...

AnotherDaniel: done

build.rs
```diff
@@ -20,10 +21,20 @@ fn main() -> std::io::Result<()> {
...
+            // cloudevent proto definitions
+            "https://raw.githubusercontent.com/cloudevents/spec/main/cloudevents/formats/cloudevents.proto", 
+
+            // uProtocol-project proto definitions
+            "https://raw.githubusercontent.com/eclipse-uprotocol/uprotocol-core-api/uprotocol-core-api-1.5.3/src/main/proto/uuid.proto",
+            "https://raw.githubusercontent.com/eclipse-uprotocol/uprotocol-core-api/uprotocol-core-api-1.5.3/src/main/proto/uri.proto",
+            "https://raw.githubusercontent.com/eclipse-uprotocol/uprotocol-core-api/uprotocol-core-api-1.5.3/src/main/proto/uattributes.proto",
+            "https://raw.githubusercontent.com/eclipse-uprotocol/uprotocol-core-api/uprotocol-core-api-1.5.3/src/main/proto/upayload.proto",
+            "https://raw.githubusercontent.com/eclipse-uprotocol/uprotocol-core-api/uprotocol-core-api-1.5.3/src/main/proto/umessage.proto",
+            "https://raw.githubusercontent.com/eclipse-uprotocol/uprotocol-core-api/uprotocol-core-api-1.5.3/src/main/proto/ustatus.proto",
```
sophokles73, line 34: in order to be able to consistently use the types generated from these protos it would be preferable to also include the other core API protos:
* file.proto
* uprotocol_options.proto
* udiscovery/v3/udiscovery.proto
* usubscription/v3/usubscription.proto
* utwin/v1/utwin.proto

stevenhartley: I'm pushing to conan now the build recipe for c++ of uprotocol-core-api, we should do the same to crates.io so you don't have to manually pull in these protos.

AnotherDaniel: done

sophokles73: I guess that would make sense. However, currently it seems that the uprotocol-core-api repo only contains archives of the whole repository and IMHO it would be more appropriate to create *release* artifacts instead, i.e. archives which only contain the proto files ...

src/lib.rs
```diff
@@ -107,62 +129,58 @@ pub mod uuid {
...
-    mod calloptions;
-    mod rpcclient;
-    mod rpcmapper;
-    mod rpcresult;
-
-    pub use calloptions::*;
-    pub use rpcclient::*;
-    pub use rpcmapper::*;
-    pub use rpcresult::*;
+pub mod uprotocol {
```
sophokles73, line 150: IMHO we are missing the module declarations for the additional proto files that you had added to the build.rs ...

AnotherDaniel: done

build.rs
AnotherDaniel: > @AnotherDaniel , if we go with @sophokles73 proposal [eclipse-uprotocol/uprotocol-core-api#76](https://github.com/eclipse-uprotocol/uprotocol-core-api/issues/76) (uprotocol-core-api only contains the protos that each language has to pull and build and push to their respective central repositories) then wouldn't it be better to pull the tar..gz, unpack and build all of them?

a) which .tar.gz - am I missing something somewhere?
b) going via an archive would pull in an additional functional dependency, and additional code into the build file; can be done, but what would be the benefit?

stevenhartley (approved):

PR #9, opened 2023-12-15
Improve UUID handling

sophokles73:
Removed the serializers in favor of implementing the corresponding `From<T>`
trait, which seems much more natural to be used in Rust than invoking explicit
helper methods. Also added matching test cases.

Removed obsolete v6 UUID variant because the uProtocol spec defines the UUIDs as being v8 only.

Fixed endianess in reading/writing bytes of uProtocol UUIDs.


src/uuid/builder/uuidbuilder.rs
```diff
@@ -183,52 +114,29 @@ mod tests {
...
     }
 
     #[test]
     fn test_uuid_for_constant_random() {
         let factory = UUIDv8Builder::new();
         let uuid1 = factory.build();
         let uuid2 = factory.build();
-        // assert_eq!(uuid1.to_fields_le().3, uuid2.to_fields_le().3); // Check that the "node" field (least significant 64 bits) is the same
         assert_eq!(uuid1.lsb, uuid2.lsb);
     }
```
PLeVasseur, line 141: Hi @sophokles73,

I'm still a newbie at Rust, so it's been fun reading through how you tackled this PR and learning from it.

I'm not sure if you want to write a bunch of unit tests at the moment, because that could make it a bit more restrictive working in the code if it's under flux, but I thought I'd take my hand at adding one here for your consideration:
```
    #[test]
    fn test_uuid_for_subsequent_generation() {
        let instant = 0x18C684468F8u64; // Thu, 14 Dec 2023 12:19:23 GMT
        let builder = UUIDv8Builder::new();
        let _uprotocol_uuid_for_instant = builder.build_with_instant(instant);
        let uprotocol_uuid_for_same_instant = builder.build_with_instant(instant);

        // same instant, version (8) and _incremented_ counter (001) should show up in UUID
        assert!(uprotocol_uuid_for_same_instant.to_hyphenated_string().starts_with("018c6844-68f8-8001-"));
        // variant is 0b10
        assert_eq!(uprotocol_uuid_for_same_instant.lsb >> 62, 0b10);
    }
```

PLeVasseur, line 141: Another one:
```rust
    #[test]
    fn test_uuid_past_max_count_in_instant() {
        let instant = 0x18C684468F8u64; // Thu, 14 Dec 2023 12:19:23 GMT
        let builder = UUIDv8Builder::new();

        // build with the same instant MAX_COUNT amount of times
        for _ in 0..=MAX_COUNT {
            let _ = builder.build_with_instant(instant);
        }

        let uprotocol_uuid_past_max_count = builder.build_with_instant(instant);

        // same instant, version (8) and _saturated_ counter (FFF) should show up in UUID
        assert!(uprotocol_uuid_past_max_count.to_hyphenated_string().starts_with("018c6844-68f8-8fff-"));
        // variant is 0b10
        assert_eq!(uprotocol_uuid_past_max_count.lsb >> 62, 0b10);
    }
```

src/uuid/builder/uuidbuilder.rs
```diff
@@ -165,15 +96,15 @@ impl UUIDv8Builder {
                 }
             } else {
                 self.msb
-                    .store((instant << 16) | (8u64 << 12), Ordering::SeqCst);
+                    .store(instant << 16, Ordering::SeqCst);
```
PLeVasseur, line 99: I'd like a little help sanity checking here. I'm trying to understand what's typical since I've not explored the internals of middleware software before.

It seems like with 48 bits, we'd be good till over 8,000 years past the Unix epoch if we count in milliseconds.

I'm aware that it's pretty ambitious to say that this protocol will exist in over 8,000 years, so would it be more typical to say, leave a comment, something to the effect of:
```rust
// 48 bits for ms since Unix epoch in roughly year 10889,  so we assume that 
// it is before then and do not check whether we have used past 48 bits
```
or would we want to be thorough and include a check for this?
```rust
// ...
// -- this is new start --
const MAX_TIMESTAMP_BITS: u8 = 48;
const MAX_TIMESTAMP_MASK: u64 = 0xffff << MAX_TIMESTAMP_BITS;
// -- this is new end --
// ...
    pub fn build_with_instant(&self, instant: u64) -> uproto_Uuid {
        let new_msb = {
            let current_msb = self.msb.load(Ordering::SeqCst);

            if instant == (current_msb >> 16) {
                if (current_msb & MAX_COUNT) < MAX_COUNT {
                    self.msb.fetch_add(1, Ordering::SeqCst);
                }
            } else {
                // -- this is new start --
                if instant & MAX_TIMESTAMP_MASK != 0 {
                    panic!("Overflowed 48 bits of unit_ts_ms")
                }
                // -- this is new end --
                self.msb
                    .store(instant << 16, Ordering::SeqCst);
            }

            self.msb.load(Ordering::SeqCst)
        };

        let mut bytes = [0u8; 16];
        bytes[..8].copy_from_slice(&new_msb.to_be_bytes());
        bytes[8..].copy_from_slice(&self.lsb);
        Builder::from_custom_bytes(bytes).into_uuid().into()
    }
}
```
and the accompanying test:
```
    #[test]
    #[should_panic(expected = "Overflowed 48 bits of unit_ts_ms")]
    fn test_uuid_past_max_unit_ts_ms () {
        // maximum value that can be stored in a 48-bit timestamp (in milliseconds)
        let max_48_bit_unix_ts_ms = (1u128 << 48) - 1;

        // add 1 millisecond to the maximum duration, to overflow
        let overflowed_48_bit_unix_ts_ms = (max_48_bit_unix_ts_ms) as u64 + 1;

        let builder = UUIDv8Builder::new();
        let _uprotocol_uuid_past_max_unix_ts_ms = builder.build_with_instant(overflowed_48_bit_unix_ts_ms);
    }
```

I suppose since we expose `UUIDv8Builder::build_with_instant()` to the user / implementor they could call it with an arbitrary `u64` that'd overflow the 48 bits we have. (Maybe a side-topic, but we'd avoid at least that issue if it were not public, but that may have other ramifications)

Given that creating UUIDs is pretty integral to uProtocol and the 8,000 years away that is, is it worth it to pay the (admittedly very small) overhead of a bitwise and for each and every UUID built?

A short not-entirely-representative example of with and without the panic on Compiler Explorer does show the extra assembly for the panic, but the times are identical when looking at the output: [Compiler Explorer example](https://godbolt.org/z/E743eET4r)

Is panicking the right thing to do in that case to indicate failure?

Sorry for writing a short story here! Hope to hear your thoughts.

src/uuid/builder/uuidbuilder.rs
PLeVasseur, line 141: I enjoyed reading over how Rust atomics are used here within the `UUIDv8Builder` to accomplish keeping a globally consistent ordering of how the UUIDs would be generated.

I wrote a unit test here to attempt to force as much contention as possible with 15 threads spun up and then waiting based on a `Barrier` till all of them are immediately before the call to `build_with_instant`.

Was cool to see it working

```rust
    use std::sync::{Arc, Barrier};
    use std::thread;

    #[test]
    fn test_uuid_atomics() {
        let thread_count_for_contention_test = 15;
        let instant = 0x18C684468F8u64; // Thu, 14 Dec 2023 12:19:23 GMT
        let builder = Arc::new(UUIDv8Builder::new());
        let barrier = Arc::new(Barrier::new(thread_count_for_contention_test));
        let mut handles = vec![];

        for _ in 0..thread_count_for_contention_test {
            let barrier_clone = barrier.clone();
            let builder_clone = Arc::clone(&builder);

            let handle = thread::spawn(move || {
                barrier_clone.wait();
                let _ = builder_clone.build_with_instant(instant);
            });
            handles.push(handle);
        }

        for handle in handles {
            handle.join().unwrap();
        }

        let uprotocol_uuid_last_one = builder.build_with_instant(instant);
        assert!(uprotocol_uuid_last_one.to_hyphenated_string().starts_with("018c6844-68f8-800f"));
    }
```

src/uuid/builder/uuidutils.rs
```diff
@@ -154,28 +114,24 @@ impl UuidUtils {
...
-                        let msb = u64::from_le_bytes(arr);
-                        let time = msb >> 16;
-                        Some(time)
-                    } else {
-                        None
-                    }
+    pub fn get_time(uprotocol_uuid: &uproto_Uuid) -> Option<u64> {
+        let uuid = uuid::Uuid::from(uprotocol_uuid);
+        match Version::from_value(uuid.get_version_num()) {
+            Some(Version::TimeOrdered) => uuid
```
PLeVasseur, line 120: I noticed [over here](https://github.com/eclipse-uprotocol/uprotocol-rust/pull/9/commits/092aea461fde0386e43be3c869e3065e7b2cbc7d#diff-dbb7b4c4d32b29138436a6c7cba0b343a3897dc209bf195e879aca12bff89161L67) the `UUIDv6Builder` was removed.

I'm curious if there's a reason that we'd still like to support retrieving the time from a v6 UUID if they're not supported over uProtocol.

Saw also in the PR summary:
> Removed obsolete v6 UUID variant because the uProtocol spec defines the UUIDs as being v8 only.

I guess we'd delete the lines:
```rust
        match Version::from_value(uuid.get_version_num()) {
            Some(Version::TimeOrdered) => uuid
                .get_timestamp()
                .map(|time| time.to_rfc4122().0),
```
if that's the case?

src/uuid/validator/uuidvalidator.rs
```diff
@@ -77,7 +69,7 @@ pub trait UuidValidator {
     /// # Errors
     ///
     /// Returns a `ValidationError` with "Invalid UUID Time" if the UUID's time component is invalid or not properly formatted.
-    fn validate_time(&self, uuid: &uproto_Uuid) -> Result<(), ValidationError> {
+    pub fn validate_time(uuid: &uproto_Uuid) -> Result<(), ValidationError> {
         if let Some(time) = UuidUtils::get_time(uuid) {
```
PLeVasseur, line 73: I suppose a side-effect of my suggestion [here](https://github.com/eclipse-uprotocol/uprotocol-rust/pull/9/files/092aea461fde0386e43be3c869e3065e7b2cbc7d#r1429467976) about removing support for UUIDv6 from `get_time()` would mean this now fails to validate.

I suppose the v6 UUID would fail to validate on the call to `validate_version()` in any case though.

src/uuid/builder/uuidbuilder.rs
sophokles73: FMPOV the handling of the next UUID being built for the same instant would be even more worth being tested :-)

src/uuid/builder/uuidbuilder.rs
sophokles73: I do share your concern regarding the potential overflow in general. 

> Is panicking the right thing to do in that case to indicate failure?

It could be, if we properly document that it does so. Panicking would have the advantage of letting client code decide if and how to make sure that the passed in value is acceptable. Some might want to check explicitly, others might be able to rely on the fact that it never does due to the way the value has been created.

On the other hand, we could also simply play it safe, change the return type to `Result<uproto_Uuid, ()> and return an Err if the passed in instant is longer than 48 bits.

FMPOV the former alternative is preferable because it preserves the simplicity of using the function (no result checking necessary). I also believe that in almost all cases, the `build()` function will be used instead, i.e. using the current time instead of a passed in instant.

WDYT?

src/uuid/builder/uuidbuilder.rs
sophokles73: while this looks interesting from a code perspective, I wonder if it actually *proves* the desired behavior or simply produces *anecdotal evidence*, given the *not very deterministic* scheduling of the threads ...

Any thought on that?

src/uuid/builder/uuidutils.rs
sophokles73: thanks for spotting :+1:

src/uuid/validator/uuidvalidator.rs
sophokles73: yes

sophokles73:
@PLeVasseur thank you very much for your review and hints for improving the PR :+1:

AnotherDaniel (approved):
Lots of simplifications, which I like. 
I don't know where the v6 UUIDs come from in the Java SDK (which they were modelled after) - maybe they are in use in legacy applications within GM, and so supporting them would be required for the Rust SDK to be an option for that context? But not very likely that this need will arise, I guess, so...

src/uuid/builder/uuidbuilder.rs
PLeVasseur: Hey, did you mean using the same builder, same instant twice?

That's what I did [over here](https://github.com/eclipse-uprotocol/uprotocol-rust/pull/9#discussion_r1428950767) in this comment.

This test was meant to check the counter saturation logic.

src/uuid/builder/uuidbuilder.rs
PLeVasseur: For my 2c, I like keeping the return type clean for the 99.9999% of cases where the timestamp would fit in 48 bits and adding some documentation that it can panic.

Perhaps the unit test I wrote in the comment could become a doc comment to illustrate when this would occur?

src/uuid/builder/uuidbuilder.rs
PLeVasseur: Yeah I thought about this a bit. Certainly the ordering that the threads are able to get access to the builder is non-deterministic in this test. The test is: "During contention among threads accessing the builder, do the atomics ensure that we have incremented the counter the appropriate number of times?"

I was considering collecting all the threads' UUIDs and confirming that we got all the counter content we were expecting, which I suppose would be a more thorough test, but opted for this "check the last counter" approach.

What do you think?

stevenhartley:
> g them would be required for the Rust SDK to be an

Long story short, UUIDv6 was the UUID format we used in production for 1.3.6 but found it wasn't efficient and carried unnecessary baggage.  We took the best of v7 and tweaked it for automotive (<1ms uniqueness, rand that is static for uE lifecycle, sortablity, etc...)

stevenhartley:
@sophokles73 we merged the 1.5 branch to main, can you re-generate this PR to main? Thanks

sophokles73:
closed in favor of #12

PR #12, opened 2023-12-19
Improve UUID handling

sophokles73:
Removed the serializers in favor of implementing corresponding From<T> traits, which seems much more natural in Rust than invoking explicit helper methods. Also added matching test cases.

Removed obsolete v6 UUID variant.

Fixed endianess in reading/writing bytes of uProtocol UUIDs.


stevenhartley:
Thanks for this Kai. You also fixed the serialize to string as the previous code was just calling to_string() and didn't put it in the formal UUID string representation form.

PR #13, opened 2023-12-20
Move helper functions to Uuid

sophokles73:
Moved relevant functions from UuidHelper and UuidValidator structs to Uuid.

Also fixed setters and getters of UResource and UAuthority to operate on &self instead of a passed in object.


stevenhartley (approved):
I don't see any issues. @dakrbo & Peter please feel free to chime in.

PLeVasseur:
Looks good to me. Consolidating functionality for UUIDs makes sense. :+1:

AnotherDaniel:
Works for me too - simplification is always good.

PR #16, opened 2023-12-21
Remove dev dependency on futures crate

sophokles73:
Also changed implementation of mapping between UPayload and Protobuf's
Any type to use explicit errors instead of default values.

### DIFF
--- a/src/rpc/rpcmapper.rs
+++ b/src/rpc/rpcmapper.rs
@@ -52,7 +52,7 @@ impl fmt::Display for RpcMapperError {
 pub struct RpcMapper;
 
 impl RpcMapper {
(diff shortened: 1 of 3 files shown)


src/rpc/rpcmapper.rs
```diff
@@ -118,96 +116,81 @@ impl RpcMapper {
     /// - is there [`UStatus`] information (transporting info about the status of an operation, sent from a remote service)?
     /// - is there payload data passed in the result, to be decoded by the caller.
     ///
-    /// This entire thing feels klunky and kludgy; this this needs to be revisited...
+    // TODO This entire thing feels klunky and kludgy; this needs to be revisited...
```
PLeVasseur, line 119: IMHO looks a bit cleaner now!

PR #17, opened 2023-12-21
Add mapping of MIME types to UPayloadFormat

sophokles73:


PLeVasseur (approved):
Seems like a useful addition. Thanks!

stevenhartley (changes requested):
I have a PR waiting to be merged where I code this in the protos so it can be fetched from generated code in lieu of coding it here.  PR should be merged tomorrow once I get +1 from Tamara.

src/proto/uprotocol/upayload.rs
```diff
@@ -14,10 +14,66 @@
...
+impl UPayloadFormat {
+    /// Gets the payload format that corresponds to a given MIME type.
+    ///
+    /// # Returns
+    ///
+    /// The corresponding payload format or [`UPayloadFormat::UpayloadFormatProtobuf`] if the MIME
+    /// type is unknown or empty.
+    pub fn from_mime_type(mime_type: &str) -> Self {
+        if let Ok(mime) = mime_type.parse::<mime::Mime>() {
+            match (mime.type_(), mime.subtype().as_str()) {
```
PLeVasseur, line 36: From what I can see the `mime` crate exposes some of the common subtypes, but doesn't appear to expose all of the ones of interest to us (e.g. `x-protobuf` is missing).

I'd like it a bit more if they did and we could rely on using e.g. `mime::X_PROTOBUF` directly.

I think this is fine tho.

Wondering if we should try to upstream more of these subtypes we're using. WDYT?

src/proto/uprotocol/upayload.rs
```diff
@@ -14,10 +14,66 @@
...
+const MIME_SUBTYPE_SOMEIP: &str = "x-someip";
+const MIME_SUBTYPE_SOMEIP_TLV: &str = "x-someip_tlv";
+const MIME_SUBTYPE_PLAIN: &str = "plain";
+
+impl UPayloadFormat {
+    /// Gets the payload format that corresponds to a given MIME type.
+    ///
+    /// # Returns
+    ///
+    /// The corresponding payload format or [`UPayloadFormat::UpayloadFormatProtobuf`] if the MIME
```
PLeVasseur, line 32: Could you elaborate a little on why we'd want to default to `UPayloadFormat::UpayloadFormatProtobuf`?

A natural default would seem to me to be `UPayloadFormat::UPAYLOAD_FORMAT_UNSPECIFIED`.

Perhaps I'm missing something.

sophokles73: My understanding is that when serializing a UMessage, a default of Protobuf encoding should be assumed, if no specific format has been set explicitly, because in 98% of all cases that will be the format being used.

@stevenhartley maybe you can shed more light on this?

src/proto/uprotocol/upayload.rs
sophokles73: In general, I share your concern. However, `application/x-protobuf` is **not** an official (i.e. registered) MIME type ...

src/proto/uprotocol/upayload.rs
PLeVasseur: Gotcha, so we hang tight till that PR merges and then we can revisit and update the logic here to match what you wrote.

sophokles73:
@PLeVasseur would you mind taking another look, now that https://github.com/eclipse-uprotocol/uprotocol-core-api/pull/89 has been merged?

PLeVasseur:
> @PLeVasseur would you mind taking another look, now that [eclipse-uprotocol/uprotocol-core-api#89](https://github.com/eclipse-uprotocol/uprotocol-core-api/pull/89) has been merged?

Sure, can do.

This might just be my personal preference, but I like when a PR is open already and there are changes if those come in another set of commits. Can help pinpoint changes that occurred since last reviewed.

In this case since there's a single commit I have to rely on my admittedly shoddy memory to recall that, haha.

WDYT? Can we try for that going forward?

build.rs
```diff
@@ -26,19 +28,19 @@ fn main() -> std::io::Result<()> {
             "https://raw.githubusercontent.com/cloudevents/spec/main/cloudevents/formats/cloudevents.proto", 
 
             // uProtocol-project proto definitions
-            "https://raw.githubusercontent.com/eclipse-uprotocol/uprotocol-core-api/uprotocol-core-api-1.5.3/src/main/proto/uuid.proto",
-            "https://raw.githubusercontent.com/eclipse-uprotocol/uprotocol-core-api/uprotocol-core-api-1.5.3/src/main/proto/uri.proto",
-            "https://raw.githubusercontent.com/eclipse-uprotocol/uprotocol-core-api/uprotocol-core-api-1.5.3/src/main/proto/uattributes.proto",
-            "https://raw.githubusercontent.com/eclipse-uprotocol/uprotocol-core-api/uprotocol-core-api-1.5.3/src/main/proto/upayload.proto",
-            "https://raw.githubusercontent.com/eclipse-uprotocol/uprotocol-core-api/uprotocol-core-api-1.5.3/src/main/proto/umessage.proto",
-            "https://raw.githubusercontent.com/eclipse-uprotocol/uprotocol-core-api/uprotocol-core-api-1.5.3/src/main/proto/ustatus.proto",
+            format!("{}/uuid.proto", UPROTOCOL_BASE_URI).as_str(),
```
PLeVasseur, line 31: Like this change 👍

src/proto/uprotocol/upayload.rs
```diff
@@ -11,13 +11,88 @@
...
 
+use mediatype::{names::*, MediaType, Name};
 use prost::Message;
 use prost_types::Any;
 
-use crate::{
-    cloudevent::serializer::SerializationError,
-    uprotocol::{Data, UPayload, UPayloadFormat},
-};
+pub use crate::types::serializationerror::SerializationError;
```
PLeVasseur, line 18: Makes sense, lets the end user make use of the error type

src/proto/uprotocol/upayload.rs
```diff
@@ -11,13 +11,88 @@
...
+const MEDIA_TYPE_TEXT_PLAIN: MediaType = MediaType::new(TEXT, PLAIN);
+
+impl UPayloadFormat {
+    /// Gets the payload format that corresponds to a given MIME type.
+    ///
+    /// # Returns
+    ///
+    /// The corresponding payload format or None if the MIME type is unsupported.
+    pub fn from_mime_type(mime_type: &str) -> Option<Self> {
+        if let Ok(mime) = MediaType::parse(mime_type) {
```
PLeVasseur, line 42: Cool, I like this, for me makes it easier to read 👍

sophokles73:
> This might just be my personal preference, but I like when a PR is open already and there are changes if those come in another set of commits. Can help pinpoint changes that occurred since last reviewed.

Absolutely, and I am totally with you. No clue why I amended the commit instead of creating another one ...

PLeVasseur:
Okay, done with review. Only one item that needs some clarification [over here](https://github.com/eclipse-uprotocol/uprotocol-rust/pull/17#discussion_r1452560282). Otherwise LGTM

src/proto/uprotocol/upayload.rs
```diff
@@ -11,13 +11,88 @@
...
+const MEDIA_TYPE_APPLICATION_SOMEIP: MediaType = MediaType::new(APPLICATION, SUBTYPE_SOMEIP);
+const MEDIA_TYPE_APPLICATION_SOMEIPTLV: MediaType = MediaType::new(APPLICATION, SUBTYPE_SOMEIP_TLV);
+const MEDIA_TYPE_TEXT_PLAIN: MediaType = MediaType::new(TEXT, PLAIN);
+
+impl UPayloadFormat {
+    /// Gets the payload format that corresponds to a given MIME type.
+    ///
+    /// # Returns
+    ///
+    /// The corresponding payload format or None if the MIME type is unsupported.
```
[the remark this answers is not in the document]
sophokles73: I do not believe that would make much sense. Let's say somebody has put in `application/vnd.my-format`. We would not be able to parse this successfully and would return `None`. However, I do not believe that an application would be able to parse the payload assuming that it is a wrapped protobuf in this case ...
When this method returns `None`, it basically means that we have no clue what the payload format is. The only case where it makes sense to assume *wrapped protobuf* is when the payload format is not set at all.
We might want to replace the returned `Option<Self>` with something like `Result<Self, ParsingError>`, in order to make it more explicit that we cannot handle the indicated format. WDYT?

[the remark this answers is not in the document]
PLeVasseur: I like the idea of explicitly noting a `ParsingError`.

Logic makes sense to me too.

Can you ping me when you've implemented the above and I'll come back?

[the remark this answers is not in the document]
sophokles73: ping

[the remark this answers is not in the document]
PLeVasseur: LGTM!

PLeVasseur (approved):
LGTM

sophokles73:
@stevenhartley can this me merged?

stevenhartley (changes requested):
Please see how I did the mime type check/conversion in https://github.com/eclipse-uprotocol/up-java/pull/72

src/proto/uprotocol/upayload.rs
```diff
@@ -11,13 +11,100 @@
...
 
-use crate::{
-    cloudevent::serializer::SerializationError,
-    uprotocol::{Data, UPayload, UPayloadFormat},
-};
+pub use crate::types::parsingerror::ParsingError;
+pub use crate::types::serializationerror::SerializationError;
+use crate::uprotocol::{Data, UPayload, UPayloadFormat};
+
+const SUBTYPE_PROTOBUF: Name = Name::new_unchecked("protobuf");
```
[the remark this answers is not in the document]
sophokles73: In theory, that would be the way to go. However, *prost-build* does not support custom options (yet). For the time being, we can either keep the const definitions in place or migrate the whole SDK to using the *rust-protobuf* crate instead of *prost-build*. *rust-protobuf* seems to have rudimentary support for extensions/custom options ...

@AnotherDaniel I had already proposed to migrate to *rust-protobuf* several weeks before ...
WDYT?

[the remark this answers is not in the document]
AnotherDaniel: As the rust-protobuf project is not dead after all, some recent commits coming in (it looked dead when I checked it out last summer), I'm good with switching to using that. Do you still have the code from that experiment lying around?

stevenhartley:
@sophokles73 you should be able to rebase this Pr and then use the EnumDescriptors like I did in the java project

sophokles73:
@stevenhartley done

stevenhartley (approved):
LGTM :-)

PR #18, opened 2023-12-25
Increase strictness / correctness of URI micro form validation

PLeVasseur:
Added helper functions on UResource and UEntity

During serialization of UURI to micro form, perform checks on
1. existence of UResource/UEntity
2. existence of id
3. id not overflowing allotted bits
4. existence of major version
5. major version not overflowing allotted bits

Also updated `is_micro_form()` to account for being false in the case that the ids are outside of expected range of 16 bits or UEntity major version it outside of expected range of 8 bits.

_Note_: Based on an issue I [raised over here](https://github.com/eclipse-uprotocol/uprotocol-rust/issues/11).


stevenhartley:
@AnotherDaniel & @sophokles73 please comment on this PR

src/proto/uprotocol/uauthority.rs
```diff
@@ -65,4 +78,51 @@ impl UAuthority {
...
+    /// # Errors
+    ///
+    /// Returns a `ValidationError` in the failure case, indicating no remote or remote is not IP
+    pub fn remote_ip_conforms(&self) -> Result<IpConformance, ValidationError> {
+        if let Some(_remote) = self.remote.as_ref() {
+            match &self.remote {
+                Some(Remote::Ip(ip)) => Ok(match ip.len() {
+                    REMOTE_IPV4_BYTES => IpConformance::IPv4,
+                    REMOTE_IPV6_BYTES => IpConformance::IPv6,
+                    _ => IpConformance::NonConformal,
```
sophokles73, line 97: Shouldn't this result in a ValidationError? my understanding is that this means we found an *ip* type remote that has neither 4 nor 16 bytes length ...

src/proto/uprotocol/uauthority.rs
```diff
@@ -65,4 +78,51 @@ impl UAuthority {
...
+                }),
+                _ => Err(ValidationError::new("Remote is not IP")),
+            }
+        } else {
+            Err(ValidationError::new("No remote"))
```
sophokles73, line 102: Is this actually an error? If I want to invoke a local service then there is no *remote*, or is there?

src/proto/uprotocol/uauthority.rs
```diff
@@ -65,4 +78,51 @@ impl UAuthority {
...
+        if let Some(_remote) = self.remote.as_ref() {
+            match &self.remote {
+                Some(Remote::Id(id)) => Ok(matches!(
+                    id.len(),
+                    REMOTE_ID_MINIMUM_BYTES..=REMOTE_ID_MAXIMUM_BYTES
+                )),
+                _ => Err(ValidationError::new("Remote is not ID")),
+            }
+        } else {
+            Err(ValidationError::new("No remote"))
```
sophokles73, line 125: see above

src/proto/uprotocol/uauthority.rs
```diff
@@ -65,4 +78,51 @@ impl UAuthority {
...
+    /// Returns whether the `Remote` `Id` meets conformance allowable range of bytes
+    ///
+    /// # Returns
+    /// Returns a `Result<bool, ValidationError>` where the error means remote is not ID or there is no remote
+    /// and happy path tells us whether the ID conforms with allowable range of bytes or not
+    ///
+    /// # Errors
+    ///
+    /// Returns a `ValidationError` in the failure case, indicating no remote or remote is not IP
+    pub fn remote_id_conforms(&self) -> Result<bool, ValidationError> {
```
sophokles73, line 115: is there really much value in returning a Result instead of just a boolean?

sophokles73 (commented):
I did not go through all of your changes. In general, I do not like the overall idea of having all the serialization and validation logic in different modules. FMPOV all of this functionality should be in the corresponding structs (thus improving cohesion) and the constraints on the fields should then be checked during creation (or deserialization) and serialization of the entities.

src/proto/uprotocol/uauthority.rs
PLeVasseur: So my thought process based on how things were laid out atm, was that in the `UriValidator::validate_micro_form()` we'd walk thru the logic in a way s.t. `remote_ip_conforms()` is only called if we know there's a remote.

Based on some of your other feedback, [like your overall review](https://github.com/eclipse-uprotocol/uprotocol-rust/pull/18#pullrequestreview-1814829206) it seems you have other thoughts.

Still gotta meditate more on your overall review to make sure I get it.

src/uri/validator/urivalidator.rs
```diff
@@ -184,21 +184,96 @@ impl UriValidator {
...
+                ))?,
+                Err(e) => Err(ValidationError::new(format!("Resource: {}", e)))?,
+            }
+        } else {
+            Err(ValidationError::new("Resource: Is missing"))?;
+        }
+
+        if let Some(authority) = uri.authority.as_ref() {
+            match authority.remote {
+                Some(Remote::Ip(_)) => {
```
PLeVasseur, line 237: Making a link

src/proto/uprotocol/uauthority.rs
PLeVasseur: I had outsourced the decision on whether non-conformal is good/bad/otherwise to the caller [like is done here](https://github.com/eclipse-uprotocol/uprotocol-rust/pull/18/files/ca15f019c2235f733944b167081bab849346c15b#r1449363340) inside of `UriValidator::validate_micro_form()`.

However, I see what you mean. I can't really imagine a scenario where the `ip` not being 4 nor 16 bytes in length would be anything other than an `ValidationError`.

src/proto/uprotocol/uauthority.rs
PLeVasseur: i didn't tackle it in this PR, but my thought process was to allow the "bubbling up" of the `ValidationError`s that had occurred, so that a caller could e.g. append all of them together on their end.

What do you think?

src/proto/uprotocol/uauthority.rs
PLeVasseur: [Yeah, same comment as over here](https://github.com/eclipse-uprotocol/uprotocol-rust/pull/18#discussion_r1449362393)

PLeVasseur:
> I did not go through all of your changes. In general, I do not like the overall idea of having all the serialization and validation logic in different modules. FMPOV all of this functionality should be in the corresponding structs (thus improving cohesion) and the constraints on the fields should then be checked during creation (or deserialization) and serialization of the entities.

First off, thanks for the review! My first Rust code review :smile: Gonna have to get this framed.

Lemme try to break the review down to make sure I'm groking it.

> In general, I do not like the overall idea of having all the serialization and validation logic in different modules. FMPOV all of this functionality should be in the corresponding structs (thus improving cohesion)

Let me try to reword it to see if I get it:

`UriValidator::validate_micro_form()` will continue to exist (since it has to get called by uP-L1 and uP-L2 APIs to validate the UURIs prior to their use), however...

its form will change to from the currently, quite large function to a much smaller one along the lines of...
```rust
    #[allow(clippy::missing_panics_doc)]
    pub fn validate_micro_form(uri: &UUri) -> Result<(), ValidationError> {
        if Self::is_empty(uri) {
            Err(ValidationError::new("URI is empty"))?;
        }

       if !UEntity::validate_micro_form(uri) {
            Err(ValidationError::new("UEntity invalid for micro form"))?;
       }

       if !UAuthority::validate_micro_form(uri) {
            Err(ValidationError::new("UAuthority invalid for micro form"))?;
       }

       if !UResource::validate_micro_form(uri) {
            Err(ValidationError::new("UResource invalid for micro form"))?;
       }

       Ok(())
    }
```

Keep in mind I'm just riffing here, so I'd probably have the `validate_micro_form()` function on each of UEntity, UResource, and UAuthority return a `ValidationError` so that I can bubble up / concatenate the reason for the error like I mentioned wanting to do [over in this comment](https://github.com/eclipse-uprotocol/uprotocol-rust/pull/18#discussion_r1449368704).

> and the constraints on the fields should then be checked during creation (or deserialization) and serialization of the entities.

Agree on the serialization bit -- the `MicroUriSerializer::serialize()` function would call `UriValidator::validate_micro_form()`, as would any implementer of `uTransport` if e.g. calling `send()`.

As for on creation, I honestly went back and forth in my mind about it. Because the structs are exposed thru the .proto files with the fields as `pub`, they could be arbitrarily constructed in any way by a user of the SDK which would not pass validation. There's really no way with our current APIs to prevent that from happening (I have thoughts on this for another day that this is not great, but this is already getting long).

I do like the idea of having builders which can help create the individual pieces in a known good fashion. I began work on, but put down for now, some work on that [over in this branch](https://github.com/PLeVasseur/uprotocol-rust_fork/tree/feature/uuri_correct_by_construction/src/uri/builder). If you think the idea has some legs, I'll work on it a bit more.

AnotherDaniel:
> I do like the idea of having builders which can help create the individual pieces in a known good fashion. I began work on, but put down for now, some work on that [over in this branch](https://github.com/PLeVasseur/uprotocol-rust_fork/tree/feature/uuri_correct_by_construction/src/uri/builder). If you think the idea has some legs, I'll work on it a bit more.

This ties in with [this discussion](https://github.com/eclipse-uprotocol/uprotocol-spec/issues/40), about reintroducing Builders for the core .proto objects. I'd be for that... and then the protoc-generated rust code could be made non-public. Thus we would hide the dependency to that code, and potential vagaries of protoc etc from the SDK users. 
Thoughts?

PLeVasseur:
I like that idea a lot!

@AnotherDaniel, @sophokles73 -- where would make sense to go from here? Close this PR and pick back up my other PR with this kinda functionality also included?

sophokles73:
@PLeVasseur FMPOV we should start with moving the checks into the corresponding structs.
Then, in an additional PR, you could start adding builders ...

PLeVasseur:
@sophokles73 -- could you take a look again to see I addressed your concern? I broke the validation logic for micro URI form up into the impls of the constituent members of a UUri.

src/proto/uprotocol/uauthority.rs
```diff
@@ -65,4 +72,53 @@ impl UAuthority {
...
+        }
+
+        if !validation_errors.is_empty() {
+            let combined_message = validation_errors
+                .into_iter()
+                .map(|err| err.to_string())
+                .collect::<Vec<_>>()
+                .join(", ");
+
+            Err(ValidationError::new(combined_message))
```
sophokles73, line 119: it doesn't look like we can end up with more than one ValidationError, can we?

src/proto/uprotocol/uentity.rs
```diff
@@ -13,8 +13,60 @@
...
+            let combined_message = validation_errors
+                .into_iter()
+                .map(|err| err.to_string())
+                .collect::<Vec<_>>()
+                .join(", ");
+
+            Err(ValidationError::new(combined_message))
+        } else {
+            Ok(())
+        }
```
sophokles73, line 70: I do not really see the value of concatenating the error messages. If we want to provide means to convey multiple error conditions then we should create a proper error struct enum with the specific error conditions and return a collection of them. 

However, IMHO we should simply return the first error we run into ...

src/proto/uprotocol/uresource.rs
```diff
@@ -25,6 +30,41 @@ impl UResource {
...
+            if id & URESOURCE_ID_VALID_BITMASK != 0 {
+                validation_errors.push(ValidationError::new(
+                    "ID does not fit within allotted 16 bits in micro form",
+                ));
+            }
+        } else {
+            validation_errors.push(ValidationError::new("ID must be present"));
+        }
+
+        if !validation_errors.is_empty() {
```
sophokles73, line 56: see previous comment

src/proto/uprotocol/uauthority.rs
PLeVasseur: Correct! Perhaps overengineered in this case. I'm fine with using `?` and bailing on early return here.

src/proto/uprotocol/uresource.rs
PLeVasseur: Yup, I agree, this could be made to bail earlier using `?`.

src/proto/uprotocol/uentity.rs
PLeVasseur: Perhaps a difference in philosophy. Putting myself in the shoes of an application developer, If I'm getting a failure, I'd like to know all the reasons.

Would you be open to me doing this:
> create a proper error struct enum with the specific error conditions and return a collection of them.
?

WDYT?

sophokles73: Sure, that would at least allow client code to actually make use of the information conveyed. On the other hand, I wonder if client code will actually ever be in a position to react to it. When it comes to supporting the debugging failures in the context of user provided URIs, I wonder if it wouldn't be more valuable to actually add logging statements to the code ...

PLeVasseur: You bring up a good point. I wonder...

I guess a use case would be if I'm trying to do a `UTransport::send(fooUuri, fooPayload, fooAttributes)`, and I get a validation error back, normally I could then attempt to call a `UDiscovery` API in order to resolve the `fooUuri`.

But perhaps if there's certain parts of the UUri that are just "bad" though, I have an alternate path where I just fail.

WDYT?

sophokles73: > normally I could then attempt to call a UDiscovery API in order to resolve the fooUuri

What API should that be? Which parts of the *bad URI* can I trust and use as the search criteria?

stevenhartley: > > normally I could then attempt to call a UDiscovery API in order to resolve the fooUuri
> 
> What API should that be? Which parts of the _bad URI_ can I trust and use as the search criteria?

The new discovery API resolveUri() https://github.com/eclipse-uprotocol/up-spec/tree/main/up-l3/udiscovery/v3#434-resolveuri was added to resolve name to numbers and vice versa

sophokles73: I can imagine that client code could use this service to resolve a (otherwise correct) URI based on names to a URI using IDs.

@PLeVasseur For that to work, it would be mandatory for the validator to provide a specific Error indicating the particular problem so that client code is not required to parse textual error messages :-)

Any thoughts regarding addition of logging statements to support debugging? I guess this would only ever be helpful during development but would most likely not be used during runtime because of the lack of file storage to log to, right?

stevenhartley: > I can imagine that client code could use this service to resolve a (otherwise correct) URI based on names to a URI using IDs.
> 
> @PLeVasseur For that to work, it would be mandatory for the validator to provide a specific Error indicating the particular problem so that client code is not required to parse textual error messages :-)
> 
> Any thoughts regarding addition of logging statements to support debugging? I guess this would only ever be helpful during development but would most likely not be used during runtime because of the lack of file storage to log to, right?

Yes. I had an example on a branch of the up-java but we are trying to figure out how to define the topics more easier in the uServices that might change the implementation but the idea is with the proto you can resolve UUris (the UEntity and/or UResource portions).

PLeVasseur: All right, well to gain some safety in the serialization to Micro form UUris, I'm going to go with @sophokles73's suggestion for now to bail at the earliest error. Let's revisit returning a more structured error explaining what's invalid when the idea is a bit more fleshed out.

src/uri/serializer/microuriserializer.rs
```diff
@@ -84,8 +84,10 @@ impl UriSerializer<Vec<u8>> for MicroUriSerializer {
     /// A `Vec<u8>` representing the serialized `UUri`.
     #[allow(clippy::cast_possible_truncation)]
     fn serialize(uri: &UUri) -> Result<Vec<u8>, SerializationError> {
-        if UriValidator::is_empty(uri) || !UriValidator::is_micro_form(uri) {
-            return Err(SerializationError::new("URI is empty or not in micro form"));
+        if let Err(validation_error) = UriValidator::validate_micro_form(uri) {
```
sophokles73, line 87: now that we return an Error we might also want to add some doc comment and maybe some examples (which can also serve as test cases)?

src/uri/validator/urivalidator.rs
```diff
@@ -1095,6 +1133,116 @@ mod tests {
...
+            .into(),
+            resource: Some(UResource {
+                id: Some(0x10000),
+                ..Default::default()
+            })
+            .into(),
+            ..Default::default()
+        };
+        let is_micro_form = UriValidator::is_micro_form(&uri);
+        assert_eq!(is_micro_form, false);
```
sophokles73, line 1153: clippy will mark this with a warning. Maybe use `assert!(!...)` instead?

src/uri/serializer/microuriserializer.rs
PLeVasseur: How might that look?

Currently have several test cases that I added to check the new functionality [down here](https://github.com/eclipse-uprotocol/up-rust/pull/18/files/daba0f5647c3923eb1385069d3b78daefeefda70#diff-200aba31079074af7b1567bf5eee9492883a3da6087d7a31d97bacf95a932209R580-R650).

sophokles73: See https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html

PLeVasseur: So you wanna see doc tests where, over on `validate_micro_form()` inside of `uurivalidator.rs`?

Or here on `serialize()`?

I think I need you to explain a bit more what you're looking for.

sophokles73: In the doc comment on the `serialize` function you could provide example code that creates a UUri, then serializes it and asserts the byte array content is as expected. That would basically be sufficient to replace the `test_serialize_uri` unit test ...
Maybe you can find some other unit tests that could serve as usage examples for the other functions as well.

PLeVasseur: Ahhh, I think I get it. You'd like to improve the docs a bit on `serialize` with some doc test examples, right?

Perhaps show 1 x success and 1 x failure?

Wanna confirm before I run off and do it :slightly_smiling_face:

sophokles73: Exactly, and while you're at it, maybe also improve the docs of some other functions as well :-)

sophokles73: And ideally, we can then remove some of the unit tests because they are now implemented as doc tests ...

PLeVasseur: All right -- gave this a go [here](https://github.com/eclipse-uprotocol/up-rust/pull/18/files#diff-50d5281e2de4d6b2de3ffb2294e4a0bb1fe84a474bd1595b080e740cba9e2aacR202-R334) and [here](https://github.com/eclipse-uprotocol/up-rust/pull/18/files#diff-200aba31079074af7b1567bf5eee9492883a3da6087d7a31d97bacf95a932209R91-R143).

Looking for any feedback you might have :slightly_smiling_face: 

BTW -- I discussed with @stevenhartley that there's a bit of ambiguity in how the id or ip or both or neither are valid states for UAuthority at the moment, so we should wait on merging this till he has it resolved in the .proto. I'll have to revise a bit based on his change to the .proto.

src/uri/serializer/microuriserializer.rs
```diff
@@ -82,10 +82,71 @@ impl UriSerializer<Vec<u8>> for MicroUriSerializer {
...
+    ///     ..Default::default()
+    ///     })
+    ///     .into(),
+    ///     ..Default::default()
+    /// };
+    /// let uprotocol_uri = MicroUriSerializer::serialize(&uri);
+    /// assert!(uprotocol_uri.is_err());
+    /// assert_eq!(
+    ///     uprotocol_uri.unwrap_err().to_string(),
+    ///     "Failed to validate micro URI format: Entity: ID does not fit within allotted 16 bits in micro form"
```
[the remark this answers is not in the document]
PLeVasseur: I'm good with detailing out various enum members to describe the errors. I can take a stab at it and you can provide feedback. WDYT?

src/uri/serializer/microuriserializer.rs
```diff
@@ -82,10 +82,71 @@ impl UriSerializer<Vec<u8>> for MicroUriSerializer {
...
+    ///     .into(),
+    ///     resource: Some(UResource {
+    ///         id: Some(29999),
+    ///     ..Default::default()
+    ///     })
+    ///     .into(),
+    ///     ..Default::default()
+    /// };
+    /// let uprotocol_uri = MicroUriSerializer::serialize(&uri);
+    /// assert!(uprotocol_uri.is_ok());
```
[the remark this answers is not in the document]
PLeVasseur: Sounds good :slightly_smiling_face:

src/uri/serializer/microuriserializer.rs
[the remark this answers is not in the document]
sophokles73: We touched on this topic already in a different context IIRC. As said there:  you should only do that if we feel that it is worth the effort. In particular, if client code will not be able to take advantage of this distinction, i.e. act accordingly during runtime when the error occurs, I do not see much value in having them. But I have no strong opinion on it ...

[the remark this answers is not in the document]
sophokles73: Yes, that is what I would suggest for the time being ...

src/uri/serializer/microuriserializer.rs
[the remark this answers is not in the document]
PLeVasseur: I added this check, but fenced it off with
```rust
#
#
```
so that the rustdoc generated example doesn't show the array literal, nor the check. IMHO helps prevent it from becoming kinda ugly :slightly_smiling_face: 

image

WDYT?

src/uri/serializer/microuriserializer.rs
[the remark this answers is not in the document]
PLeVasseur: Updated to align w/ your feedback

PLeVasseur:
> You got the idea 👍 I made some suggestions for improvement. WDYT?

Alrighty, I updated based on your suggestions. Can you take another look?

Cargo.toml
```diff
@@ -50,7 +50,7 @@ uuid = { version = "1.7", features = ["v8"] }
 [build-dependencies]
 protobuf-codegen = { version = "3.3" }
 protoc-bin-vendored = { version = "3.0" }
-ureq = { version = "2.7" }
+ureq = { version = "2.9", features = ["proxy-from-env"] }
```
PLeVasseur, line 53: @sophokles73, @AnotherDaniel -- I snuck this one in here to allow me to build the project (namely the `ureq` download of the `.proto`s) behind our wonderful corporate proxy :rofl:

src/uri/validator/urivalidator.rs
```diff
@@ -186,19 +185,218 @@ impl UriValidator {
...
+    ///     })
+    ///     .into(),
+    ///     resource: Some(UResource {
+    ///         id: Some(29999),
+    ///         ..Default::default()
+    ///     })
+    ///     .into(),
+    ///     ..Default::default()
+    /// };
+    /// # assert_eq!(uri.clone().authority.unwrap().id.unwrap(), expected_id_bytes);
```
sophokles73, line 285: what is the point of this and having the expected_id_bytes?

src/uri/validator/urivalidator.rs
```diff
@@ -186,19 +185,218 @@ impl UriValidator {
...
+    ///     resource: Some(UResource {
+    ///         id: Some(0x10000), // <- exceeds allotted 16 bits
+    ///         ..Default::default()
+    ///     })
+    ///     .into(),
+    ///     ..Default::default()
+    /// };
+    /// let val_micro_form = UriValidator::validate_micro_form(&uri);
+    /// assert!(val_micro_form.is_err());
+    /// ```
```
sophokles73, line 357: is there also an example showing a successful validation? Maybe as the first example?

src/uri/serializer/microuriserializer.rs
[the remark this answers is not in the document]
sophokles73: I don't see your point. is this a formatting problem?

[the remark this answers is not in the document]
PLeVasseur: Oh... it's not a problem. Guess a taste thing? I didn't really wanna have a huge vec literal crowd out the main point of the example. Hope that makes sense.

src/uri/validator/urivalidator.rs
PLeVasseur: Oh yeah, this is the bit from the rustdoc book about [hiding parts of your doctest](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example) which may not be relevant or they key thing you wanna showcase.

src/uri/validator/urivalidator.rs
PLeVasseur: Yeah, I could add a successful validation :+1:

src/uri/serializer/microuriserializer.rs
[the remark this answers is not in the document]
sophokles73: I cannot see any *fenced off literal check* anywhere around here. This is the *Serialize* function, so I would have expected to see an example where serialization actually works and is being asserted by comparing the returned value with an (expected) byte array. However, there seem to be only examples where serialization fails ...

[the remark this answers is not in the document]
PLeVasseur: Sorry, I misread. Will add in this check!

src/uri/validator/urivalidator.rs
PLeVasseur: OK -- I'll remove this, sorry I misread! :slightly_smiling_face:

src/uri/validator/urivalidator.rs
```diff
@@ -186,19 +185,218 @@ impl UriValidator {
...
-    /// Returns `true` if the URI contains numbers, allowing it to be serialized into micro format.
+    /// Returns `Ok(())` if the URI contains numbers which will fit in the allotted space,
+    /// allowing it to be serialized into micro format.
+    ///
+    /// # Errors
+    ///
+    /// Otherwise returns `ValidationError` containing description of error.
+    ///
+    /// # Examples
+    ///
```
PLeVasseur, line 203: Need to add a happy path success case here

PLeVasseur:
Hey folks -- sorry I missed that the up-core-api change got merged fixing the ambiguity around id/ip. I'll have a go at updating this to unblock #28

cc @sophokles73 @AnotherDaniel

PLeVasseur:
> Hey folks -- sorry I missed that the up-core-api change got merged fixing the ambiguity around id/ip. I'll have a go at updating this to unblock #28

Even better, @AnotherDaniel's PR got in first! Makes life easy.

Could you take a look and see if I hit the points we discussed the other day @sophokles73?

.gitignore
```diff
@@ -13,4 +13,5 @@ Cargo.lock
 # MSVC Windows builds of rustc generate these, which store debugging information
 *.pdb
 
-.vscode/
\ No newline at end of file
+.vscode/
+.idea/
```
sophokles73, line 17: maybe you can add an EOL here?

src/uri/serializer/microuriserializer.rs
```diff
@@ -268,7 +327,7 @@ mod tests {
         assert!(uprotocol_uri.is_err());
         assert_eq!(
             uprotocol_uri.unwrap_err().to_string(),
-            "URI is empty or not in micro form"
+            "Failed to validate micro URI format: URI is empty"
```
sophokles73, line 330: I'd rather remove these literal comparisons as they are hard to maintain. In particular, when not using constants for the messages ...

PLeVasseur: Yeah, I could go through this file and remove these.

To a degree I was following along with the rest of the unit tests.

You want me to remove them all from this file?

PLeVasseur: Okay, went ahead and did that, could you take another look?

sophokles73:
@stevenhartley can you approve the workflow run so that we can see that the changes pass all checks? And then merge ...

PR #19, opened 2023-12-29
Update uListener type for supporting Zenoh

evshary:
1. Add SYNC trait for listener callback
2. Add result argument for listener callback and users can be notified if something is wrong.
3. Fix some clippy issues

### DIFF
--- a/src/transport/datamodel/utransport.rs
+++ b/src/transport/datamodel/utransport.rs
@@ -59,7 +59,7 @@ pub trait UTransport {
     async fn register_listener(
         &self,
         topic: UUri,
-        listener: Box<dyn Fn(UMessage) + Send + Sync + 'static>,
+        listener: Box<dyn Fn(Result<UMessage, UStatus>) + Send + Sync + 'static>,
     ) -> Result<String, UStatus>;
 
     /// Unregister a listener for a given topic. Messages arriving on this topic will no longer be processed
--- a/src/transport/datamodel/utransport.rs
+++ b/src/transport/datamodel/utransport.rs
@@ -59,7 +59,7 @@ pub trait UTransport {
     async fn register_listener(
         &self,
         topic: UUri,
-        listener: Box<dyn Fn(UMessage) + Send + 'static>,
+        listener: Box<dyn Fn(UMessage) + Send + Sync + 'static>,
     ) -> Result<String, UStatus>;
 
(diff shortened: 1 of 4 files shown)


PLeVasseur:
Hi @evshary -- could you please ensure you're signed the ECA?

I think you can click the `Details` text there on workflow check and then click `Sign` at the bottom to do so, if you already have an account with the Eclipse Foundation.

src/transport/datamodel/utransport.rs
```diff
@@ -52,14 +52,18 @@ pub trait UTransport {
...
     /// # Returns
-    /// Asynchronously returns a `Result<String, UStatus>`. On success, returns a `String` containing an identifier that can be used for unregistering the listener later. On failure, returns `Err(UStatus)` with the appropriate failure information.
+    /// Asynchronously returns a `Result<String, UStatus>`.
+    /// On success, returns a `String` containing an identifier that can be used for unregistering the listener later.
+    /// On failure, returns `Err(UStatus)` with the appropriate failure information.
     async fn register_listener(
         &self,
         topic: UUri,
-        listener: Box<dyn Fn(UMessage) + Send + 'static>,
+        listener: Box<dyn Fn(Result<UMessage, UStatus>) + Send + Sync + 'static>,
```
PLeVasseur, line 66: It would be good to hear from @sophokles73 on this as he wanted a chance to see practical implications.

evshary:
> Hi @evshary -- could you please ensure you're signed the ECA?
> 
> I think you can click the `Details` text there on workflow check and then click `Sign` at the bottom to do so, if you already have an account with the Eclipse Foundation.

Hi @PLeVasseur , I've already signed the ECA, but not sure why it didn't pass yesterday (I tried several time)
I guess maybe it's related to some database synchronization issue.
Anyway, I just tried again, and it works now.

src/transport/datamodel/utransport.rs
evshary: Maybe I should also link to the issue https://github.com/eclipse-uprotocol/uprotocol-rust/issues/14 for better tracing

sophokles73: I have been able to apply this change to my code without any problems. However, this might be due to the fact that my listener is quite trivial and is thus not very prone to be affected by this additional constraint.

sophokles73 (approved):
LGTM

PLeVasseur (approved):

PR #23, opened 2024-01-23
Use rust-protobuf instead of prost

sophokles73:
The library now uses the rust-protobuf crate instead
of prost to generate structs from the uProtocol proto files.

All of the library's code has been adapted to the changed API
while trying to maintain the library's existing external behavior.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -11,94 +11,99 @@
  * SPDX-License-Identifier: Apache-2.0
  ********************************************************************************/
 
-use prost_build::Config;
 use std::env;
 use std::fs;
 use std::path::Path;
+use std::path::PathBuf;
 
-fn main() -> std::io::Result<()> {
-    // use vendored protoc instead of relying on user provided protobuf installation
-    std::env::set_var("PROTOC", protoc_bin_vendored::protoc_bin_path().unwrap());
+const UPROTOCOL_BASE_URI: &str = "https://raw.githubusercontent.com/eclipse-uprotocol/uprotocol-core-api/uprotocol-core-api-1.5.5/uprotocol";
 
-    if let Err(err) = get_and_build_protos(
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    get_and_build_protos(
         &[
             // cloudevent proto definitions
             "https://raw.githubusercontent.com/cloudevents/spec/main/cloudevents/formats/cloudevents.proto", 
+        ],
+        "cloudevents",
+    )?;
 
+    get_and_build_protos(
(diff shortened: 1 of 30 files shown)


build.rs
```diff
@@ -11,94 +11,99 @@
...
-use prost_build::Config;
 use std::env;
 use std::fs;
 use std::path::Path;
+use std::path::PathBuf;
 
-fn main() -> std::io::Result<()> {
-    // use vendored protoc instead of relying on user provided protobuf installation
-    std::env::set_var("PROTOC", protoc_bin_vendored::protoc_bin_path().unwrap());
+const UPROTOCOL_BASE_URI: &str = "https://raw.githubusercontent.com/eclipse-uprotocol/uprotocol-core-api/uprotocol-core-api-1.5.4/uprotocol";
```
stevenhartley, line 19: @sophokles73 please pull 1.5.5, I added a fix for uAuthority

stevenhartley (changes requested):
Please update to pull in 1.5.5 of core-api, other than that I have no issue

src/uuid/builder/uuidbuilder.rs
```diff
@@ -16,7 +16,7 @@ use std::convert::Into;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::{SystemTime, UNIX_EPOCH};
 
-use crate::uprotocol::Uuid as uproto_Uuid;
+use crate::uprotocol::uuid::UUID as uproto_Uuid;
```
PLeVasseur, line 19: One thing I thought of here... looking to hear your thoughts. I don't know what the convention is, but if I'm looking for impls of UUid or its usage within this crate, I think it makes it a little harder to find due to this `use foo as bar`.

WDYT? Other than that, I'm good with the changes.

PLeVasseur (commented):
LGTM

sophokles73:
@stevenhartley I have upgraded to uProtocol 1.5.5 now ...

stevenhartley:
@sophokles73 could you check this PR? I might have broke it while trying to merge.

sophokles73:
@stevenhartley no problem, I will rebase and squash my branch. I recommend that you do not use *merge commits* when merging a PR but use *rebasing* instead. This will provide for a much cleaner and easier to follow commit log.

src/uuid/builder/uuidbuilder.rs
sophokles73: I have removed the alias in the latest commit I have pushed ...

PR #25, opened 2024-01-25
Add GitHub workflow for creating 3rd party dependency list

sophokles73:
Adresses #39

A GitHub workflow has been added which determines these by means of the [Eclipse Dash license tool](https://github.com/eclipse/dash-licenses).

DO NOT REVIEW YET, THIS IS AN EXPERIMENT FOR NOW

### DIFF
--- a/.github/workflows/check-dependencies.yaml
+++ b/.github/workflows/check-dependencies.yaml
@@ -35,7 +35,7 @@ jobs:
     - name: cargo tree
       working-directory: ${{github.workspace}}
       run: |
-        cargo tree -e normal --prefix none --no-dedupe \
+        cargo tree -e no-build,no-dev --prefix none --no-dedupe \
           | sort -u \
           | grep -v '^[[:space:]]*$' \
           | grep -v up-rust \
--- a/.github/workflows/check-dependencies.yaml
+++ b/.github/workflows/check-dependencies.yaml
@@ -38,7 +38,7 @@ jobs:
         cargo tree -e normal --prefix none --no-dedupe \
           | sort -u \
           | grep -v '^[[:space:]]*$' \
-          | grep -v uprotocol-sdk \
+          | grep -v up-rust \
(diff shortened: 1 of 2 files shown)


sophokles73:
My overall intention is to have the following behavior:

* The `build and test` workflow should not require all 3rd party deps to be checked, i.e. it should succeed if the code compiles, the tests succeed and the linter does not complain.
* The `check 3rd party deps` workflow should **succeed** if all deps have been vetted successfully. In that case, there is no need to keep a record of the checks.
* The `check 3rd party deps` workflow should **fail**, if any of the deps cannot be vetted successfully. In that case, we should keep a record of the output created by the Dash license tool as an attachment to the workflow run. Having the workflow fail in this case makes it easier to detect that we need to work on the deps, because we can easily spot the failed runs based on their visual markup on GitHub and the fact that we receive a mail for the failed run.

@AnotherDaniel @stevenhartley @PLeVasseur
Does this make sense to you?

PLeVasseur:
> @AnotherDaniel @stevenhartley @PLeVasseur Does this make sense to you?

Yup, seems reasonable to me :slightly_smiling_face:

SebastianSchildt:
For the second step - piping the dash input file through dash, we do have an Action 

https://github.com/eclipse-kuksa/kuksa-actions/tree/main/check-dash

here is an example how to use it https://github.com/eclipse/kuksa.val/blob/cc2e0a67af497685cbcb3d92700f85359f8eab87/.github/workflows/kuksa_databroker_build.yml#L239

And here https://github.com/eclipse/kuksa.val/actions/runs/8230232480 is an example of the outputs, see "Dash data" artefact, as well as "Bill of Material Check summary"

it can't do Dash ticket creation (yet).

Feel free to use from our repo, or just "copy & paste" it over here, it is OSS after all :).  I would just assume this pattern will be used in many uproto repos, so likely a bit of reusability would be good (which is how we ended up with a common action)

sophokles73:
@AnotherDaniel I'd like to start using the check now ...

AnotherDaniel (approved):
LGTM

PR #26, opened 2024-01-25
Update GitHub actions depending on deprecated node.js

sophokles73:
The GitHub workflows have produced a warning message indicating that some of the GitHub Actions used in the workflow had been using a deprecated version of Node.js and should thus be updated ...


sophokles73:
@stevenhartley there is no change in the functionality of the workflow here ...

PR #27, opened 2024-01-29
Update RpcClient::invoke_method() to return a UMessage in support of symmetry / flexibility

PLeVasseur:
Hey there :wave: 

As I worked on the uStreamer we figured out we'll need RpcClient::invoke_method() to return a UMessage in order to make the operations performed by the uStreamer upon the CEs coming in and out as generic as possible.

@stevenhartley put out a [spec change](https://github.com/eclipse-uprotocol/up-spec/issues/61) and made the [change over in up-java](https://github.com/eclipse-uprotocol/up-java/pull/74) already. More details can be understood from there.

PR #28, opened 2024-01-30
Cleanup and pull in recent Java SDK features

AnotherDaniel:
- clean up and update main readme
- remove contributors list (not needed, have git logs)
- pull in changes and additions from Java SDK PR #60

### DIFF
--- a/src/proto/cloudevents/protocloudevent.rs
+++ b/src/proto/cloudevents/protocloudevent.rs
@@ -23,6 +23,77 @@ use crate::cloudevents::cloud_event::CloudEventAttributeValue;
 use crate::cloudevents::cloud_event::Data as CloudEventData;
 use crate::cloudevents::CloudEvent as CloudEventProto;
 
+use crate::uprotocol::umessage::UMessage;
+use crate::uprotocol::{SerializationError, UPriority};
+use crate::uri::serializer::{LongUriSerializer, UriSerializer};
+
+impl TryFrom<UMessage> for cloudevents::Event {
+    type Error = SerializationError;
+
+    fn try_from(source_event: UMessage) -> Result<Self, Self::Error> {
+        let attributes = source_event.attributes.get_or_default();
+        let source = attributes.source.get_or_default();
+        let payload = source_event.payload.get_or_default();
+        let uri = LongUriSerializer::serialize(source)?;
+
+        if attributes.id.is_none() {
+            return Err(SerializationError::new("Empty attributes ID"));
+        }
+        if attributes.type_.enum_value().is_err() {
+            return Err(SerializationError::new("Bad attributes type"));
+        }
+
+        let event_builder = cloudevents::EventBuilderV10::new()
+            .id(attributes.id.get_or_default())
+            .ty(attributes.type_.enum_value_or_default().to_type_string())
+            .source(uri);
+        let mut event = event_builder
+            .build()
+            .map_err(|e| SerializationError::new(format!("Error creating cloudevent: {e}")))
+            .unwrap();
+
+        let ctype = payload.format.enum_value_or_default();
+        event.set_datacontenttype(ctype.to_media_type());
+
+        if payload.has_value() {
+            event.set_data_unchecked(payload.value().to_vec());
+        }
+        if let Some(ttl) = attributes.ttl {
+            event.set_extension("ttl", ExtensionValue::Integer(ttl as i64));
+        }
+        if let Some(token) = &attributes.token {
+            event.set_extension("token", ExtensionValue::String(token.to_string()));
+        }
+        if attributes
+            .priority
+            .enum_value()
+            .is_ok_and(|p| p != UPriority::UPRIORITY_UNSPECIFIED)
+        {
+            event.set_extension(
+                "priority",
+                ExtensionValue::String(
+                    attributes.priority.enum_value().unwrap().to_priority_code(),
+                ),
+            );
+        }
+        if let Some(sink) = &attributes.sink.clone().into_option() {
+            let uri = LongUriSerializer::serialize(sink)?;
+            event.set_extension("sink", ExtensionValue::String(uri));
+        }
+        if let Some(commstatus) = attributes.commstatus {
+            event.set_extension("commstatus", ExtensionValue::Integer(commstatus as i64));
+        }
+        if let Some(reqid) = &attributes.reqid.clone().into_option() {
+            event.set_extension("reqid", ExtensionValue::String(reqid.to_string()));
+        }
+        if let Some(plevel) = attributes.permission_level {
+            event.set_extension("plevel", ExtensionValue::Integer(plevel as i64));
+        }
+
+        Ok(event)
+    }
+}
+
 impl From<CloudEventProto> for cloudevents::Event {
     fn from(source_event: CloudEventProto) -> Self {
         let mut subject: Option<String> = None;
@@ -124,7 +195,6 @@ impl From<cloudevents::Event> for CloudEventProto {
         let mut ext_list = HashMap::<String, CloudEventAttributeValue>::new();
 
         // subject
-        // TODO how is this serialized by eg the Java libraries, considering cloudevent.proto is missing dedicated attributes for this?
         if let Some(subject) = source_event.subject() {
             let s = CloudEventAttributeValue {
                 attr: Some(Attr::CeString(subject.to_string())),
@@ -134,7 +204,6 @@ impl From<cloudevents::Event> for CloudEventProto {
         }
 
         // timestamp
-        // TODO how is this serialized by eg the Java libraries, considering cloudevent.proto is missing dedicated attributes for this?
         if source_event.time().is_some() {
             let time = *source_event.time().unwrap();
             let sys_time: std::time::SystemTime = time.into();
@@ -149,7 +218,6 @@ impl From<cloudevents::Event> for CloudEventProto {
         }
 
         // dataschema
-        // TODO how is this serialized by eg the Java libraries, considering cloudevent.proto is missing dedicated attributes for this?
         if let Some(schema) = source_event.dataschema() {
             let ds = CloudEventAttributeValue {
                 attr: Some(Attr::CeUri(schema.to_string())),
@@ -159,7 +227,6 @@ impl From<cloudevents::Event> for CloudEventProto {
         }
 
         // contenttype
-        // TODO how is this serialized by eg the Java libraries, considering cloudevent.proto is missing dedicated attributes for this?
         if let Some(contenttype) = source_event.datacontenttype() {
             let ct = CloudEventAttributeValue {
                 attr: Some(Attr::CeString(contenttype.to_string())),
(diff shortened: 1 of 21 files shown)


src/proto/cloudevents/protocloudevent.rs
```diff
@@ -23,6 +23,58 @@ use crate::cloudevents::cloud_event::CloudEventAttributeValue;
 use crate::cloudevents::cloud_event::Data as CloudEventData;
 use crate::cloudevents::CloudEvent as CloudEventProto;
 
+use crate::uprotocol::umessage::UMessage;
+use crate::uri::serializer::{LongUriSerializer, UriSerializer};
```
PLeVasseur, line 27: IMHO it'd be nice if we're only using `LongUriSerializer` to only have to use that one, i.e. only have to:
```rust
use crate::uri::serializer::{LongUriSerializer};
```

I think we can achieve that by [modifying this line](https://github.com/eclipse-uprotocol/up-rust/blob/ae311dd9d170d7030344fb020a9ed83342f6b0d6/src/uri/serializer/longuriserializer.rs#L19):
```rust
use crate::uri::serializer::{SerializationError, UriSerializer};
```
to
```rust
pub use crate::uri::serializer::{SerializationError, UriSerializer};
```

src/proto/cloudevents/protocloudevent.rs
```diff
@@ -23,6 +23,58 @@ use crate::cloudevents::cloud_event::CloudEventAttributeValue;
...
+
+impl From<UMessage> for cloudevents::Event {
+    fn from(source_event: UMessage) -> Self {
+        let source = source_event.source.get_or_default();
```
PLeVasseur, line 31: I noticed leaning on getting default and unwraps throughout.

I'm wondering if we should be doing any additional error checking and make it TryFrom instead to reflect that it may fail?

WDYT?

src/rpc/rpcclient.rs
```diff
@@ -13,37 +13,39 @@
...
+    /// Returns a `RpcClientResult` which contains the response message.
+    /// If the invocation fails, it contains a `UStatus` detailing the failure reason.
     async fn invoke_method(
         &self,
-        topic: UUri,
-        payload: UPayload,
-        attributes: UAttributes,
+        method: UUri,
+        request: UPayload,
+        options: CallOptions,
```
PLeVasseur, line 49: :+1:

src/rpc/rpcmapper.rs
```diff
@@ -122,8 +122,8 @@ impl RpcMapper {
     ///
     // TODO This entire thing feels klunky and kludgy; this needs to be revisited...
     pub fn map_response_to_result(response: RpcClientResult) -> RpcPayloadResult {
-        let payload = response?; // Directly returns in case of error
-        Any::try_from(payload)
+        let message = response?; // Directly returns in case of error
+        Any::try_from(message.payload.unwrap())
```
PLeVasseur, line 126: Funnily enough, I was doing the same thing [over in this PR](https://github.com/eclipse-uprotocol/up-rust/pull/27) :laughing: 

We should probably be a little more careful here, [similar to what I did on that PR](https://github.com/eclipse-uprotocol/up-rust/pull/27/files#diff-8d66726b55938f3831ad9ba37cb66af6c321e6091a883099cde61b586056cbc3R86) s.t. we don't accidentally panic.

src/rpc/rpcmapper.rs
```diff
@@ -296,22 +296,33 @@ mod tests {
...
     use crate::cloudevents::CloudEvent as CloudEventProto;
-    use crate::uprotocol::UMessageType;
+    use crate::uprotocol::{UMessage, UMessageType, UPayload};
 
     fn build_status_response(code: UCode, msg: &str) -> RpcClientResult {
-        let status = UStatus::fail_with_code(code, msg);
+        let status: UStatus = UStatus::fail_with_code(code, msg);
         let any = RpcMapper::pack_any(&status)?;
-        Ok(any.try_into().unwrap())
+        let payload =
```
PLeVasseur, line 307: Again, funny we both did this haha

Kai gave some good feedback on that PR on how to handle this nicely, perhaps you could [reference that](https://github.com/eclipse-uprotocol/up-rust/pull/27/files#diff-8d66726b55938f3831ad9ba37cb66af6c321e6091a883099cde61b586056cbc3R130-R136)?

src/uri/validator/urivalidator.rs
```diff
@@ -198,7 +216,23 @@ impl UriValidator {
...
+    /// evaluates whether a given `UAuthority` meets the criteria to be considered as a micro format authority.
+    ///
+    /// # Arguments
+    ///
+    /// * `authority` - A reference to the `UAuthority` to check.
+    ///
+    /// # Returns
+    ///
+    /// Returns `true` if the `UAuthority` can be represented in micro format, `false` otherwise.
+    pub fn is_micro_form_authority(authority: &UAuthority) -> bool {
```
PLeVasseur, line 234: This one I would pause on tho. IMHO it makes more sense to merge the [PR I've got going](https://github.com/eclipse-uprotocol/up-rust/pull/18) for this functionality. That PR goes into a lot more detail in checking also that the fields from the UUri are able to fit the specified number of bits & bytes available.

If the PR I have for checking micro form goes through, that can then be mirrored back to `up-java`.

PLeVasseur:
Overall good to stay in sync with `up-java`, but I'd recommend removing [this bit here for `is_micro_form_authority()`](https://github.com/eclipse-uprotocol/up-rust/pull/28#discussion_r1471866303) since [this other PR](https://github.com/eclipse-uprotocol/up-rust/pull/18) I have does further checking into buffer size & available number of bytes.

If you could remove that bit, I think we're good :+1:

stevenhartley (changes requested):
Mapping error for priority that needs fixing

src/proto/cloudevents/protocloudevent.rs
AnotherDaniel: Good catch!

src/rpc/rpcmapper.rs
AnotherDaniel: Pulled that in - and apologies for waltzing over your work there!

src/uri/validator/urivalidator.rs
AnotherDaniel: Removed!

src/proto/cloudevents/protocloudevent.rs
AnotherDaniel: Addendum - this doesn't really work for some reason, dependent code still won't compile without explicitly use-ing the Trait implementation. I guess I'll revert this change for now.

src/proto/cloudevents/protocloudevent.rs
```diff
@@ -23,6 +23,58 @@ use crate::cloudevents::cloud_event::CloudEventAttributeValue;
...
+        if payload.has_value() {
+            event.set_data_unchecked(payload.value().to_vec());
+        }
+        if let Some(ttl) = attributes.ttl {
+            event.set_extension("ttl", ExtensionValue::Integer(ttl as i64));
+        }
+        if let Some(token) = &attributes.token {
+            event.set_extension("token", ExtensionValue::String(token.to_string()));
+        }
+        if let Ok(priority) = attributes.priority.enum_value() {
```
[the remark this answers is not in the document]
AnotherDaniel: I made the changes to achieve this - not yet based off the referenced PR, but manually. Will change that bit as soon as up-core-api#98 is merged.

src/proto/cloudevents/protocloudevent.rs
PLeVasseur: My bad! Looks like this is not possible in Rust. We have to bring the trait into scope explicitly with a `use`. Seems this is a conscious design decision to help when multiple traits define things with the same names and helps readability to understand where a function could be coming from.

AnotherDaniel: heh - forget all that, found the problem. We will have a try_from.

src/proto/cloudevents/protocloudevent.rs
AnotherDaniel: we have a try_from now - However, next I will have to spend some quality time revisiting the test setup for cloudevent.rs, stay tuned for another change in that area.

AnotherDaniel:
Will take a couple of days yet - need to revisit some test cases at the very least, plus want to wait for up-core-api 1.5.6
Will let you know once this is ready!

stevenhartley:
Yes I will do a release this week, not waiting for the other changes that I want to review at the generic uProtocol meeting on Thursday.

src/transport/datamodel/utransport.rs
```diff
@@ -41,7 +51,7 @@ pub trait UTransport {
     ///
     /// # Returns
     /// Returns () on success, otherwise an Err(UStatus) with the appropriate failure information.
-    async fn send(
+    async fn send_parts(
```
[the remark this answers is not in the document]
AnotherDaniel: Agree and fixed

src/transport/datamodel/utransport.rs
```diff
@@ -47,7 +47,7 @@ pub trait UTransport {
     /// On success, returns a tuple containing the received `UPayload` and `UAttributes` wrapped in `Ok(_)`.
     /// If receiving the message fails, it returns `Err(UStatus)`,
     /// where `UStatus` contains a `UCode` indicating the specific error or failure reason.
-    async fn receive(&self, topic: UUri) -> Result<(UPayload, UAttributes), UStatus>;
+    async fn receive(&self, topic: UUri) -> Result<UMessage, UStatus>;
```
sophokles73, line 50: The doc comment still refers to the original types. Which also is a good example for why you should not mention the types of parameters in the doc comments at all (they are obvious from the generated documentation anyway) ;-)

sophokles73:
@AnotherDaniel There is no doubt that the changes in this PR are relevant and helpful :+1:
However, the PR contains quite a few changes which do not necessarily seem to be related to each other. This makes it quite hard to review and rebase, because it touches on so many areas. For the future, I would suggest to try to keep the scope of a PR small and consistent and try to only change things that are in scope. That will make it much easier to get things reviewed and merged.

AnotherDaniel:
I'm still mentally working my way down from the all-encompassing initial commit, but point taken :)

AnotherDaniel:
@PLeVasseur , I had to do a number on the MicroUriSerializer code, to make it work with up-core-api v1.5.6  -   sorry to get in the way of of your #18 there, probably you can update that one to the same up-core-api and then take over my changes?
@sophokles73 I'm on the fence about the remaining UAuthority::get_name()

sophokles73:
Why do you want to remove the getters from UAuthority in the first place? Aren't they useful anymore?

AnotherDaniel:
They now come with the generated code...

sophokles73:
@AnotherDaniel can you rebase this PR?

PLeVasseur:
Hey folks -- sorry I missed that the up-core-api change got merged fixing the ambiguity around id/ip. I'll have a go at updating this.

cc @sophokles73 @AnotherDaniel

sophokles73 (approved):
LGTM

AnotherDaniel:
@stevenhartley the changes you requested wrt up-core-api 1.5.6 (umessage option strings) are in now

PR #33, opened 2024-02-08
Add examples to UPayload doc comments

sophokles73:


src/proto/uprotocol/upayload.rs
```diff
@@ -25,6 +25,19 @@ impl UPayloadFormat {
     /// # Errors
     ///
     /// Returns an error if the given string is not a valid media type string or is unsupported by uProtocol.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
```
PLeVasseur, line 31: The docs tell us that you needn't put the `rust` there, by default anything fenced in code block ticks will be interpreted as Rust code. :+1:

PR #34, opened 2024-02-08
Add convenience functions and examples to UMessageType

sophokles73:
Added functions for mapping message types to/from string encoded integers and type strings. These should be helpful when encoding the message type to a property of a transport layer PDU.

Also removed the implementation for Display which might have been mistaken to produce a UMessageType's stable string encoding.


PLeVasseur (approved):
LGTM :+1: 

Lovin' all the doctests :slightly_smiling_face:

stevenhartley (approved):

PR #37, opened 2024-02-09
String-Uri conversions can now fail

AnotherDaniel:
String-Uri conversions can now fail, instead of silently returning (empty) default objects.

### DIFF
--- a/src/proto/uprotocol/uuri.rs
+++ b/src/proto/uprotocol/uuri.rs
@@ -11,19 +11,19 @@
  * SPDX-License-Identifier: Apache-2.0
  ********************************************************************************/
 
-use crate::uprotocol::uri::UUri as uproto_Uuri;
+use crate::uprotocol::uri::UUri;
 use crate::uprotocol::SerializationError;
 use crate::uri::serializer::{LongUriSerializer, MicroUriSerializer, UriSerializer};
 
-impl TryFrom<uproto_Uuri> for String {
+impl TryFrom<UUri> for String {
     type Error = SerializationError;
 
-    fn try_from(value: uproto_Uuri) -> Result<Self, Self::Error> {
+    fn try_from(value: UUri) -> Result<Self, Self::Error> {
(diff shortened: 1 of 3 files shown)


src/proto/uprotocol/uuri.rs
```diff
@@ -12,44 +12,37 @@
  ********************************************************************************/
 
 use crate::uprotocol::uri::UUri as uproto_Uuri;
+use crate::uprotocol::SerializationError;
```
sophokles73, line 15: Can we remove the alias for the `UUri` type? It doesn't seem to conflict with anything else, does it?

src/uri/serializer/uriserializer.rs
```diff
@@ -41,22 +41,22 @@ pub trait UriSerializer<T> {
...
     /// Builds a fully resolved `UUri` from the serialized long format and the serialized micro format.
     ///
     /// # Arguments
-    /// * `long_uri` - `UUri` serialized as a string.
-    /// * `micro_uri` - `UUri` serialized as a byte slice.
+    /// * `long_uri` - uri serialized as a string.
+    /// * `micro_uri` - uri serialized as a byte slice.
     ///
     /// # Returns
     /// Returns an `Option<UUri>` object serialized from one of the forms. Returns `None` if the URI
```
sophokles73, line 62: FMPOV we should change this function to return a Result instead of an Option to make it more explicit that this actually constitutes an erroneous situation.

AnotherDaniel: Good point, done

PLeVasseur (approved):
Nice, using TryFrom and modelling the errors more explicitly 👍

sophokles73:
@AnotherDaniel do you want to also include moving the serializer code to UUri in this PR or do you want to do that in a separate PR?

sophokles73:
Then I would propose to do it in a separate PR and merge this PR now.

sophokles73 (approved):
LGTM

PR #38, opened 2024-02-09
Remove uuid crate from UUID's public API

sophokles73:
Also removed obsolete functions, added doc comments and refactored unit tests into documentation tests.

### DIFF
--- a/src/proto/uprotocol/uuid.rs
+++ b/src/proto/uprotocol/uuid.rs
@@ -11,7 +11,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ********************************************************************************/
 
-use std::str::FromStr;
+use std::{hash::Hash, str::FromStr};
 
 use crate::uprotocol::UUID;
 
@@ -187,6 +187,13 @@ impl UUID {
     /// ```
     pub fn is_uprotocol_uuid(&self) -> bool {
         self.is_custom_version() && self.is_rfc_variant()
+    }
+}
+
+impl Hash for UUID {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        let bytes = (self.msb, self.lsb);
+        bytes.hash(state)
     }
 }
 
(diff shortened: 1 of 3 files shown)


AnotherDaniel:
Another dependency we don't need another, I like it. Can this also already include the change to Cargo.toml, removing the uuid crate from the list there?

sophokles73:
No we cannot remove it, we still need the `uuid` crate for serializing and deserializing to/from a hyphenated string. I have just *hidden* our usage of it by not using `uuid::Uuid` in our public API anymore.

src/proto/uprotocol/uuid.rs
```diff
@@ -37,140 +41,205 @@ impl std::fmt::Display for UuidConversionError {
...
-
-        if uuid.get_variant() != Variant::RFC4122 {
-            return false;
-        }
-
-        true
-    }
-}
-
-impl From<uproto_Uuid> for Uuid {
```
PLeVasseur, line 97: IIRC this was kinda handy to have since the proto `UUID` was not hashable.

I [used it over here](https://github.com/PLeVasseur/uprotocol-platform-linux-zenoh/blob/65468148fa49389b8cdc56083490f11e8da4b3e0/ustreamer/src/plugins/up_client_full.rs#L372) in some prototype code for the uStreamer.

Is there any way to either keep this or an alternative we can do within `up-rust` to support hashing the `UUID`s?

sophokles73: I guess all we need to do is implement the `Hash` trait for `UUID`. I will add that ...

sophokles73:
@PLeVasseur does this work for you?

PLeVasseur:
> @PLeVasseur does this work for you?

Yeah, this looks like what I need. Thanks :slightly_smiling_face:

PLeVasseur (approved):
> @PLeVasseur does this work for you?

Yeah, this looks like what I need. Thanks :slightly_smiling_face:

sophokles73:
@stevenhartley can we merge this?

PR #41, opened 2024-02-15
Rename UUIDv8Builder to UUIDBuilder

sophokles73:
Since we are only using v8 UUIDs in uProtocol, there is no more
reason to add the version to the name of the builder.


stevenhartley (approved):

PLeVasseur (approved):
Yup, this makes sense :+1:

sophokles73:
@stevenhartley can you merge this?

PR #44, opened 2024-02-16
Fix build

sophokles73:
The required feature for configuring a proxy from ENV variables
is available starting from version 2.9.1.


sophokles73:
@PLeVasseur would you mind checking if this works for you, approve and then ping @stevenhartley  to merge?

stevenhartley (approved):
I pulled the PR, ran build, test, and doc targets like the workflow and it all builds cleanly so approving. Thanks Kai!

PLeVasseur:
> The required feature for configuring a proxy from ENV variables is available starting from version 2.9.1.

Ah shoot. Thanks for catching this.

Issue #43, opened 2024-02-15
Replace UAttributesBuilder with UMessageBuilder


sophokles73:
With the changes introduced to `UTransport::send`, it seems more desirable to be able to easily build a `UMessage` to be sent instead of creating `UAttributes`.

PLeVasseur:
Makes sense. Doing this for the different types of `UMessageType`, right? (Request, Response, Publish...) And we'd also pass in the `UPayload` to the builder?

sophokles73:
Exactly

sophokles73:
Closed via #46

PR #46, opened 2024-02-20
Replace UAttributesBuilder with UMessageBuilder

sophokles73:
This is for #43


sophokles73:
@PLeVasseur @AnotherDaniel WDYT?

PLeVasseur:
Will take a look today :+1:

src/transport/builder/umessagebuilder.rs
```diff
@@ -0,0 +1,733 @@
...
+    comm_status: Option<i32>,
+    request_id: Option<&'a UUID>,
+    payload: Option<Bytes>,
+    payload_format: UPayloadFormat,
+}
+
+impl<'a> Default for UMessageBuilder<'a> {
+    fn default() -> Self {
+        UMessageBuilder {
+            validator: Box::new(crate::transport::validator::PublishValidator),
```
PLeVasseur, line 44: I was struggling with understanding how much of this we want to match the spec or be opinionated atm.

It seemed to me reading the spec for [`UTransport::send()`](https://github.com/eclipse-uprotocol/up-spec/tree/main/up-l1#211-parameters)
> SHOULD use the Miro Uri serialization format

I realize that this is `SHOULD`, but it's at least the 'strongest' terms used for what a UUri should comply with. I see that PublishValidator [currently checks for ability to serialize down to Long format](https://github.com/SoftwareDefinedVehicle/uprotocol-sdk-rust_fork/blob/4570552d76ac8ed823dc8b68d0211f63c9229119/src/cloudevent/validator/cloudeventvalidator.rs#L355C48-L355C59), but IMHO it should do so for Micro format instead.

WDYT?

PLeVasseur: Also tagging in @stevenhartley for some feedback, as I know, for example the Android Binder implementation _does not_ support all the numeric portions of the UUri. So if we said that the numeric portions are required only, that'd be a bit of a leak that'd let us build a message which is unable to be routed over Android Binder.

PLeVasseur: I opened [an issue](https://github.com/eclipse-uprotocol/up-spec/issues/68) over on `up-spec` asking a few questions. I think the answer to question 1. is relevant to this PR.

sophokles73: > I see that PublishValidator [currently checks for ability to serialize down to Long format](https://github.com/SoftwareDefinedVehicle/uprotocol-sdk-rust_fork/blob/4570552d76ac8ed823dc8b68d0211f63c9229119/src/cloudevent/validator/cloudeventvalidator.rs#L355C48-L355C59), but IMHO it should do so for Micro format instead.

IMHO the idea is to make sure that it is serializable to *at least* long format, which sounds reasonable to me. If a particular transport wants to use micro format, then it can still first try to serialize to micro form and could then either fall back to long format or fail sending the message altogether.

src/transport/builder/umessagebuilder.rs
```diff
@@ -0,0 +1,733 @@
...
+    ///
+    /// # Returns
+    ///
+    /// A message ready to be sent using [`crate::transport::datamodel::UTransport::send`].
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the properties set on the builder do not represent a valid set of [`UAttributes`].
+    pub fn build(&self) -> Result<UMessage, ValidationError> {
+        let uuid_builder = UUIDBuilder::new();
```
PLeVasseur, line 528: IMHO this is a bit against the spirit of what is described as the [UUID in the spec](https://github.com/eclipse-uprotocol/up-spec/blob/ceb40c3e218ba29241f8cc8f10aa141a7a2d3c84/basics/uuid.adoc#2-specification), in particular w/r/t `rand_b`:
> * MUST be a 62 bits pseudo-random number generated per uE
> * MUST NOT change during the lifecycle of the uE
In other words, there should not ever be a time that a uE should be coming up with a new `rand_b` to describe itself on the fly and the `build_with()` function below should be the only available way to obtain a final `UMessage`.

WDYT?

src/transport/builder/umessagebuilder.rs
```diff
@@ -0,0 +1,733 @@
...
+    /// ```
+    pub fn request(
+        method_to_invoke: &'a UUri,
+        reply_to_address: &'a UUri,
+        ttl: u32,
+    ) -> UMessageBuilder<'a> {
+        UMessageBuilder {
+            validator: Box::new(crate::transport::validator::RequestValidator),
+            message_type: UMessageType::UMESSAGE_TYPE_REQUEST,
+            source: Some(reply_to_address),
```
PLeVasseur, line 169: Also related to the [issue I made](https://github.com/eclipse-uprotocol/up-spec/issues/68).

src/transport/builder/umessagebuilder.rs
```diff
@@ -0,0 +1,733 @@
...
+    /// ```
+    pub fn response(
+        reply_to_address: &'a UUri,
+        request_id: &'a UUID,
+        invoked_method: &'a UUri,
+    ) -> UMessageBuilder<'a> {
+        UMessageBuilder {
+            validator: Box::new(crate::transport::validator::ResponseValidator),
+            message_type: UMessageType::UMESSAGE_TYPE_RESPONSE,
+            source: Some(invoked_method),
```
PLeVasseur, line 217: Also related to the [issue I made](https://github.com/eclipse-uprotocol/up-spec/issues/68).

src/transport/builder/umessagebuilder.rs
```diff
@@ -0,0 +1,733 @@
...
+    /// assert_eq!(response_message.attributes.sink, Some(reply_to_address).into());
+    /// assert_eq!(response_message.attributes.reqid, Some(request_id).into());
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn response_for_request(request_attributes: &'a UAttributes) -> UMessageBuilder<'a> {
+        UMessageBuilder {
+            validator: Box::new(crate::transport::validator::ResponseValidator),
+            message_type: UMessageType::UMESSAGE_TYPE_RESPONSE,
+            source: request_attributes.sink.as_ref(),
```
PLeVasseur, line 263: Also related to the [issue I made](https://github.com/eclipse-uprotocol/up-spec/issues/68).

src/transport/builder/umessagebuilder.rs
```diff
@@ -0,0 +1,733 @@
...
+    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// let topic = UUri::try_from("my-vehicle/cabin/1/doors.driver_side#status")?;
+    /// let message = UMessageBuilder::publish(&topic)
+    ///                 .with_payload("locked".into(), UPayloadFormat::UPAYLOAD_FORMAT_TEXT)
+    ///                 .build()?;
+    /// assert!(message.payload.is_some());
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn with_payload(
```
PLeVasseur, line 297: WDYT about offering another function as well, e.g. `with_protobuf_payload()` where we put a trait bound on the input parameter `payload` s.t. it must be serializable using the `protobuf` crate?

I think that could be handy to offer that service to within the function do the serialization as well as an option.

src/transport/builder/umessagebuilder.rs
PLeVasseur: But why _at least_ long format? IIRC the `up-client-sommr-rust` client library will probably only speak over Micro Uri format.

src/transport/builder/umessagebuilder.rs
sophokles73: I had the same thought when I added the `build_with` function. If I understand you correctly, you propose to replace the original `build` function with `build_with` (and rename that to `build` then), right?

src/transport/builder/umessagebuilder.rs
sophokles73: I had thought about that as well. We already have types from the `protobuf` crate in the public API so it shouldn't be an issue.

PLeVasseur: Okay. Darn. Yeah, let's leave it for now.

PLeVasseur: OK. Here's an alternative proposal. What about making a `build_with_payload()` which also takes an input that is bound to be serializable via `Protobuf` returns a `Result`?

Maybe a stretch, but that'd let us allow it to fail, without it being in the middle of the chain.

WDYT?

sophokles73: Not sure if I understand. Can you provide some example code?

PLeVasseur: Sure! Something like the below. I didn't compile it, so I may have missed a thing or two. Only thing is that returning a ValidationError in that case seems a little odd.

May also be possible to factor out some of the common code (e.g. reqid, attributes stuff) between this and the `build()` we talked about turning `build_with()` into.

WDYT?

```rust
/// Creates the message based on the builder's state, including a `Profobuf` serializable object.
///
/// # Arguments
///
/// `uuid_builder` - A (shared) builder for creating UUIDs.
/// `payload` - A payload which is serializable using `Protobuf`
///
/// # Returns
///
/// A message ready to be sent using [`crate::transport::datamodel::UTransport::send`].
///
/// # Errors
///
/// Returns an error if the properties set on the builder do not represent a valid set of [`UAttributes`] or
/// there is an issue in serializing the payload using `Protobuf`
pub fn build_with_payload<T: Message>(&self, uuid_builder: &UUIDBuilder, payload: T) -> Result<UMessage, ValidationError> {
        let reqid = self.request_id.cloned().or_else(|| {
            if self.message_type == UMessageType::UMESSAGE_TYPE_REQUEST {
                Some(uuid_builder.build())
            } else {
                None
            }
        });
        let attributes = UAttributes {
            id: Some(uuid_builder.build()).into(),
            type_: self.message_type.into(),
            source: self.source.cloned().into(),
            sink: self.sink.cloned().into(),
            priority: self.priority.into(),
            ttl: self.ttl,
            token: self.token.cloned(),
            permission_level: self.permission_level,
            commstatus: self.comm_status,
            reqid: reqid.into(),
            ..Default::default()
        };
        let buf = payload
            .write_to_bytes()
            .map_err(|_e| ValidationError::new("Failed to serialize payload to bytes"))?;
        self.validator.validate(&attributes).map(|_| {
            let payload = buf
                .as_ref()
                .map(|bytes| Some(Data::Value(bytes.to_vec())))
                .map(|data| UPayload {
                    format: self.payload_format.into(),
                    data,
                    ..Default::default()
                });
            UMessage {
                attributes: Some(attributes).into(),
                payload: payload.into(),
                ..Default::default()
            }
        })
    }
}
```

src/transport/builder/umessagebuilder.rs
```diff
@@ -0,0 +1,733 @@
...
+    ///
+    /// A message ready to be sent using [`crate::transport::datamodel::UTransport::send`].
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the properties set on the builder do not represent a valid set of [`UAttributes`].
+    pub fn build_with(&self, uuid_builder: &UUIDBuilder) -> Result<UMessage, ValidationError> {
+        let reqid = self.request_id.cloned().or_else(|| {
+            if self.message_type == UMessageType::UMESSAGE_TYPE_REQUEST {
+                Some(uuid_builder.build())
```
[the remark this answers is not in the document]
sophokles73: I was also thinking about removing `with_request_id` and instead add it as a parameter to the `request` factory function. WDYT?

src/transport/builder/umessagebuilder.rs
sophokles73: Ah, ok, now I see it. This will need to be the final step then. If we want to go that way then I would propose to change the `with_payload` function to `build_with_payload` accordingly. WDYT?

PLeVasseur: Yup, makes sense!

PLeVasseur (approved):
LGTM. Examples are great.

PR #49, opened 2024-02-21
Refactor LongUriSerializer into Rust-idiomatic TryFrom trait implementation

AnotherDaniel:
This change involves many changed lines, but covers only one topic: get rid of the not-very-Rusty serialization concept for LongUris, and replace with a lot more idiomatic TryFrom Uri<->String implementation. This also involved refactoring and streamlining all of the related Uri test cases.


AnotherDaniel:
@PLeVasseur I'll do the same to MicroUriSerialization, now that your big improvement PR is merged

src/uri/serializer/uriserializer.rs
```diff
@@ -66,8 +66,10 @@ pub trait UriSerializer<T> {
             return Err(SerializationError::new("Input uris are empty"));
         }
 
-        let long_uri = UUri::try_from(long_uri)?;
-        let micro_uri = UUri::try_from(micro_uri.to_vec())?;
+        let long_uri =
+            UUri::try_from(long_uri).map_err(|e| SerializationError::new(e.to_string()))?;
+        let micro_uri = UUri::try_from(micro_uri.to_vec())
+            .map_err(|e| SerializationError::new(e.to_string()))?;
```
sophokles73, line 72: IMHO we should use different names for the variables here in order to prevent confusion with the passed in parameters ...

AnotherDaniel: Done

src/proto/uprotocol/uuri.rs
```diff
@@ -11,38 +11,571 @@
...
+    ///     })
+    ///     .into(),
+    ///     authority: None.into(),
+    ///     ..Default::default()
+    /// };
+    ///
+    /// let uri_from = String::try_from(uri).unwrap();
+    /// assert_eq!("/example.com//rpc.raise", uri_from);
+    /// ````
+    fn try_from(uri: UUri) -> Result<Self, Self::Error> {
```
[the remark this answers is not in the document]
AnotherDaniel: Done

src/proto/uprotocol/uuri.rs
```diff
@@ -11,38 +11,617 @@
...
+
+        let uri = if let Some(index) = uri.find(':') {
+            uri[index + 1..].to_string()
+        } else {
+            uri.replace('\\', "/")
+        };
+        let is_local: bool = !uri.starts_with("//");
+        let uri_parts = Self::pattern_split(&uri, "/");
+
+        if uri_parts.len() < 2 {
```
PLeVasseur, line 161: I'm reading through the UUri spec now to parse out what's going on here. May be nice to add a little bit of explanatory comments.

e.g. here could be
```
// If we don't have at a minimum a UEntity and UResource, we fail
```
or something to that effect

WDYT?

src/proto/uprotocol/uuri.rs
```diff
@@ -11,38 +11,617 @@
...
+    ///
+    /// let uri_from = UUri::from_str("/example.com//rpc.raise").unwrap();
+    /// assert_eq!(uri, uri_from);
+    /// ````
+    fn from_str(uri: &str) -> Result<Self, Self::Err> {
+        if uri.is_empty() {
+            return Err(SerializationError::new("URI is empty"));
+        }
+
+        let uri = if let Some(index) = uri.find(':') {
```
PLeVasseur, line 153: Similar comment as [over here](https://github.com/eclipse-uprotocol/up-rust/pull/49/files#r1498107883).

Could maybe put
```rust
// strip leading `up` and `:`
```

src/proto/uprotocol/uuri.rs
```diff
@@ -11,38 +11,617 @@
...
+            if uri_parts.len() > 2 {
+                version = uri_parts[2].to_string();
+            }
+            if uri_parts.len() > 3 {
+                resource = Some(UResource::from(uri_parts[3].as_str()));
+            }
+        } else {
+            if uri_parts.len() > 2 {
+                if uri_parts[2].trim().is_empty() {
+                    return Err(SerializationError::new("URI is invalid"));
```
PLeVasseur, line 182: I think here the error message could be made a bit more specific, like this:
```rust
return Err(SerializationError::new("Remote URI missing UAuthority"));
```
WDYT?

src/proto/uprotocol/uuri.rs
PLeVasseur: Or perhaps even better, the error message could be updated to reflect this e.g.
```rust
return Err(SerializationError::new("URI missing UEntity or UResource"));
```

src/proto/uprotocol/uuri.rs
```diff
@@ -11,38 +11,617 @@
...
+
+        // Compatibility note: in the Java SDK, UEntity versions are 'int', therefore default to 0. For some reason,
+        // UUris with a 0 version, the version is not properly serialized back to a string (0 is omitted). Anyways,
+        // we handle this properly. There either is a version, or there is not.
+        let mut ve: Option<u32> = None;
+        if !version.is_empty() {
+            if let Ok(version) = version.parse::<u32>() {
+                ve = Some(version);
+            } else {
+                return Err(SerializationError::new("URI is invalid"));
```
PLeVasseur, line 213: Similar thought here to update the returned error to be more descriptive.

src/proto/uprotocol/uuri.rs
```diff
@@ -11,38 +11,617 @@
...
+    ///
+    /// # Arguments
+    /// * `long_uri` - uri serialized as a string.
+    /// * `micro_uri` - uri serialized as a byte slice.
+    ///
+    /// # Returns
+    /// If successful, returns an UUri object serialized from the input formats. Returns `SerializationError` if the deserialization
+    /// fails or the resulting uri cannot be resolved.
+    pub fn build_resolved(long_uri: &str, micro_uri: &[u8]) -> Result<UUri, SerializationError> {
+        if long_uri.is_empty() && micro_uri.is_empty() {
```
PLeVasseur, line 302: Maybe should be `||` instead with updated corresponding message noting one of them is empty _or_ alternatively (maybe better?) check each one independently so that the most descriptive error message can be returned, noting which one is empty.

Why? Well, I guess if either one of them is empty, we can't say that we're returning a fully resolved URI.

src/proto/uprotocol/uuri.rs
```diff
@@ -11,38 +11,617 @@
...
+            return Err(SerializationError::new("Input uris are empty"));
+        }
+
+        let long_uri = UUri::from_str(long_uri)?;
+        let micro_uri = UUri::try_from(micro_uri.to_vec())?;
+
+        let mut auth = micro_uri.authority.unwrap_or_default();
```
PLeVasseur, line 309: I suppose the idea here is that we would have failed and bailed out due to the two lines prior where we attempt to deserialize, so it's safe to assume that we have the appropriate values available for these three lines. Is that right?

If so, I wouldn't mind some comment to that effect, but WDYT?

PLeVasseur: I guess in general I get a little uncomfortable pulling out the `unwrap_or_default()` unless there's clear preconditions that prevent us from _actually picking the default_.

What about using `ok_or_else` something like this?
```rust
let mut auth = micro_uri.authority.ok_or_else(|| SerializationError::new("Missing UAuthority");
let mut ue = micro_uri.entity.ok_or_else(|| SerializationError::new("Missing UEntity");
let mut ure = long_uri.resource.ok_or_else(|| SerializationError::new("Missing UResource");
```
It'd make me a little more comfortable if there's some refactoring in the future to guard against that possibility. Or maybe I'm being paranoid, haha.

WDYT?

src/proto/uprotocol/uuri.rs
```diff
@@ -11,38 +11,617 @@
...
+    ///     })
+    ///     .into(),
+    ///     authority: None.into(),
+    ///     ..Default::default()
+    /// };
+    ///
+    /// let uri_from = String::try_from(&uri).unwrap();
+    /// assert_eq!("/example.com//rpc.raise", uri_from);
+    /// ````
+    fn try_from(uri: &UUri) -> Result<Self, Self::Error> {
```
PLeVasseur, line 86: Making a link

src/proto/uprotocol/uuri.rs
```diff
@@ -11,38 +11,617 @@
...
+            entity: Some(ue).into(),
+            resource: Some(ure).into(),
+            ..Default::default()
+        };
+
+        if UriValidator::is_resolved(&uri) {
+            Ok(uri)
+        } else {
+            Err(SerializationError::new(format!(
+                "Could not resolve uri {uri}"
```
PLeVasseur, line 336: If we do this, wouldn't we only print the long form via [impl TryFrom<&UUri> for String](https://github.com/eclipse-uprotocol/up-rust/pull/49/files#r1498198835)?

Would it be preferable to use the `Debug` output here instead `{:?}` so that we can read from the error all the contents of the UUri, including the numeric portions?

WDYT?

src/proto/uprotocol/uuri.rs
AnotherDaniel: Picking the second option :)

src/proto/uprotocol/uuri.rs
AnotherDaniel: Not arguing against your point - but should this behavior also be revised in [up-java](https://github.com/eclipse-uprotocol/up-java/blob/b502e03da95648c21e8f8b8daffe0ea082d69e82/src/main/java/org/eclipse/uprotocol/uri/serializer/UriSerializer.java#L66), then?

src/proto/uprotocol/uuri.rs
AnotherDaniel: Sure, let's put that it. If it gets on peoples nerves, we can always reduce verbosity again

src/proto/uprotocol/uuri.rs
AnotherDaniel: Valid points - added more error reporting

src/proto/uprotocol/uuri.rs
PLeVasseur: Yeah, from my point of view `up-java` is not gospel, the spec is. Perhaps you could open an easy issue on `up-spec` confirming.

For example, from what I saw the more intensive checking on Micro URIs is not in `up-java`, but probably should be. I should go open an issue on that repo regarding this actually.

PLeVasseur (approved):
LGTM. Thx for updating to add more explanations 👍

src/proto/uprotocol/uuri.rs
```diff
@@ -11,38 +11,634 @@
...
+        })
     }
 }
 
 impl TryFrom<&str> for UUri {
     type Error = SerializationError;
 
-    fn try_from(value: &str) -> Result<Self, Self::Error> {
-        LongUriSerializer::deserialize(value.into())
+    /// Attempts to serialize a `String` into a `UUri`.
```
sophokles73, line 239: how about also implementing `TryFrom<String>`?

AnotherDaniel: Done

sophokles73 (approved):
LGTM

sophokles73:
@AnotherDaniel do you need to squash this or can I merge it *as is*?

AnotherDaniel:
> @AnotherDaniel do you need to squash this or can I merge it _as is_?

Already done - should be good to merge!

PR #52, opened 2024-02-25
Fix RPC response URI validator

evshary:
According to the spec, `validateRpcResponse(UUri): UStatus` shoud return false while `UUri.resource.name != "rpc" || UUri.resource.instance != "response" || UUri.resource.id != 0`
https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc#3-validators


src/uri/validator/urivalidator.rs
```diff
@@ -163,9 +163,9 @@ impl UriValidator {
                     .as_ref()
                     .map_or(false, |instance| instance.contains("response"));
 
-                let has_non_zero_id = resource.id.map_or(false, |id| id != 0);
+                let has_zero_id = resource.id.map_or(false, |id| id == 0);
 
-                return has_valid_instance || has_non_zero_id;
+                return has_valid_instance && has_zero_id;
```
PLeVasseur, line 168: Agreed, this is a good catch!

src/uri/validator/urivalidator.rs
```diff
@@ -163,9 +163,9 @@ impl UriValidator {
                     .as_ref()
                     .map_or(false, |instance| instance.contains("response"));
```
PLeVasseur, line 164: I guess technically the spec says it should be equal to, not that it should contain "response". Shouldn't we check for equality?

Tagging @AnotherDaniel in here to sanity check me.

PLeVasseur, line 164: Also noticed that we're not checking for `name` to be "rpc", unlike within `is_rpc_method()`, even though the spec says to check the name is "rpc" for both validators.

Am I reading that right? Can you check me here again @AnotherDaniel?

AnotherDaniel: You're correct. The 'contains' is a leftover from me dumbly doing a 1:1 convert from up-java, but the spec requires equality.

evshary: Maybe I'm wrong, but I guess it's intended to check "rpc" by using `is_rpc_method` [here](https://github.com/eclipse-uprotocol/up-rust/pull/52/files#diff-50d5281e2de4d6b2de3ffb2294e4a0bb1fe84a474bd1595b080e740cba9e2aacR159).

evshary: I've updated it.

PLeVasseur: [Spec says](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc#3-validators) we should check for it:
image

evshary: Updated.

AnotherDaniel (approved):
LGTM

PLeVasseur (approved):
LGTM

src/uri/validator/urivalidator.rs
```diff
@@ -1145,11 +1146,7 @@ mod tests {
             name: "hartley".into(),
             ..Default::default()
         };
-        let resource = UResource {
-            name: "rpc".into(),
-            id: Some(19999),
-            ..Default::default()
-        };
+        let resource = UResourceBuilder::for_rpc_response();
```
sophokles73, line 1149: can we please keep the code that explicitly specifies the (valid) resource field values?

sophokles73 (approved):
LGTM

Issue #31, opened 2024-02-08
Reorg and cleanup of source structure


AnotherDaniel:
Hi fellow up-rustaceans, 
I want to do some fundamental cleanup and reorg of the SDK source file structure. Lining out the planned changes here, for feedback and also so we can maybe agree on a hands-off period for a week or so, where I then do all of this without making merging of parallel work more nightmarish than it needs to be.

The core things I want to do: 

- clean out the src/proto folder (this is there for 'historical reasons', but the introduction of up-core-api was the final nail in the coffin for the original idea)
- make the cloudevents bits an optional crate feature 

More detailed, I'm planning to do the following moves:

- the top-level uprotocol logic (mostly in /proto and the \*/datamodel folders atm) into the top-level /src folder
- the /uuid, /uri etc folders will remain, but contain the trait definitions for their associated types (validators, serializers etc)
- /rpc stays as it is
- /types only contains error definitions at this point, these well me merged and moved to top-level

__Note__: the external interface of the up-rust crate wouldn't change, so for users this should be transparent (with the possible exception of the cloudevents feature becoming optional, which I'd put into it's own PR though)

Any comments? Do any of you have some large changes in the pipeline that this should be time-synced with? Otherwise, I'd wait for the current batch of PRs to get merged, then pull of the above changes, then everyone could get back to work.

PLeVasseur:
> * make the cloudevents bits an optional crate feature

I like this idea a lot :heart_eyes: 

> Otherwise, I'd wait for the current batch of PRs to get merged, then pull of the above changes, then everyone could get back to work.

Sounds good to me. I just have the one PR related to to Micro Form Uri that's there.

There's another PR I am working on to add the ability to compile & use `up-rust` as an Android module, but this will take time and I don't think it needs to block your work.

AnotherDaniel:
Update: as soon as #54 and #52 are merged, I'll do the big file-move as described above.

AnotherDaniel:
Starting the big shuffle now...

@sophokles73 @evshary @PLeVasseur contrary to what I thought initially, I'm beginning to think I'd like to change the external crate interface, basic simplifications like:

```use up_rust::uprotocol::uri::SerializationError;```

becomes

```use up_rust::uri::SerializationError;```


Would you kill me if I did a bit of that?

sophokles73:
I wouldn't :-)

PLeVasseur:
Makes sense to me =^)

[...]

PR #57, opened 2024-03-11
Declaration of cloudevents feature

AnotherDaniel:
As announced in #31, this is making the cloudevents dependency an optional up-rust crate feature.


src/lib.rs
```diff
@@ -17,23 +17,28 @@
...
+//! * `uuid` module, which generates and validates UUIDs as per the uProtocol specification
 //!
 //! For user convenience, all of these modules export their types on up_rust top-level, except for (future) optional features.
 //!
+//! ## Optional features
+//!
+//! Some crate features are made optional, these include:
+//!
+//! * `cloudevents` feature, offering a way to represent uProtocol messages using the [cloudevents data model](https://cloudevents.io/)
+//!
```
PLeVasseur, line 41: I'm not sure how "Rust-literate" we're expecting users to be. If we're expecting them to be fairly literate, maybe my comment not needed. Perhaps here we could provide a link to how to enable an optional feature in Rust or provide specifically how to enable that for this crate?

WDYT?

AnotherDaniel: Hm, maybe not in the crates.io description - but I'll add a bit of text to the project README, wdyt?

README.md
```diff
@@ -8,27 +8,28 @@ The module contains the factory methods, serializers, and validators for all dat
...
+Building the library is as simple as running `cargo build` in the project root directory. The build pulls [uprotocol protobuf definitions](https://github.com/eclipse-uprotocol/up-core-api) that constitute the core types used in the library. All of these are compiled automatically by `build.rs`. The resulting Rust code is made available to the project via the `up_core_api::` internal module.
 
-__Note:__ the library current uses protobuf message definitions from the cloudevents project. Specifically, the build process downloads:
+__Note:__ the library uses non-stable features from the uuid crate, notably version 8 UUIDs. These features are defined in `cargo.toml` (where the uuid crate dependency is declared), and require a compiler flag to be included in the build. This is configured in `.cargo/config.toml`.
+
+### Optional features
+
+The up-rust crate contains some optional features, as per the create documentation (refer to src/lib.rs). One of these features is the usage of protobuf message definitions from the cloudevents project. Specifically, with this feature enabled the build process downloads:
```
PLeVasseur, line 19: `create` => `crate`?

README.md
```diff
@@ -8,27 +8,28 @@ The module contains the factory methods, serializers, and validators for all dat
...
 
-__Note:__ the library uses non-stable features from the uuid crate, notably version 8 UUIDs. These features are defined in `cargo.toml` (where the uuid crate dependency is declared), and require a compiler flag to be included in the build. This is configured in `.cargo/config.toml`.
+```json
+{
+    "rust-analyzer.cargo.features": "all"
+}
+```
 
-### Using the library
+For more details on Rust (cargo) features, please refer to [The Cargo Book](https://doc.rust-lang.org/cargo/reference/features.html).
```
PLeVasseur, line 31: :heart:

src/rpc/rpcmapper.rs
```diff
@@ -354,6 +359,7 @@ mod tests {
         Ok(message)
     }
 
+    #[cfg(feature = "cloudevents")]
```
PLeVasseur, line 362: I haven't used feature-gating in Rust very much yet. Did you test this locally to make sure these tests are run w/ the feature on and not when the feature is off?

Cargo.toml
```diff
@@ -33,11 +33,17 @@ repository = "https://github.com/eclipse-uprotocol/up-rust"
 rust-version = "1.66"
 version = "0.1.5"
 
+[features]
+cloudevents = ["dep:cloudevents-sdk"]
```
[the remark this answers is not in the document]
AnotherDaniel: true

Cargo.toml
```diff
@@ -48,7 +48,7 @@ mediatype = "0.19"
 protobuf = { version = "3.3" }
 rand = { version = "0.8" }
 regex = { version = "1.10" }
-serde = { version = "1.0", features = ["derive"] }
+serde = { version = "1.0", features = ["derive"], optional = true }
```
sophokles73, line 51: `serde_json` is also required by the cloudevent-sdk but is not really required by `uurivalidator::tests`. In fact, there is a test which reads in a JSON file that contains valid and invalid URIs ... IMHO this should be replaced by `#[test_case]`s to remove the (unnecessary) dependency on `serde_json`.

.github/workflows/lint-and-test.yaml
```diff
@@ -59,7 +59,7 @@ jobs:
       - name: Run tests and report code coverage
         run: |
           # enable nightly features so that we can also include Doctests
-          RUSTC_BOOTSTRAP=1 cargo tarpaulin -o xml -o lcov -o html --doc --tests
+          RUSTC_BOOTSTRAP=1 cargo tarpaulin -o xml -o lcov -o html --doc --tests --all-features
```
sophokles73, line 62: now we no longer test if it actually compiles/works _without_ the `cloudevents` feature, do we?

AnotherDaniel: This is test code coverage collection - makes most sense to run with all features.
I'd say we need something dedicated in addition now, to cover the matrix of possible features. Am looking into that atm...

PLeVasseur: I'll take it you'll ping us when you have a solution for that in mind for another round of code review. :+1:

Issue #55, opened 2024-03-01
Should we keep UAttributesBuilder?


evshary:
I love the PR #46, which makes the code cleaner.
However, I found that we still need to create the UAttributes by ourselves inside `invoke_method` while sending the RPC request.

```rust
async fn invoke_method(
    &self,
    method: UUri,
    request: UPayload,
    options: CallOptions,
) -> RpcClientResult;
```

Now I use `UMessageBuilder` to create the UMessage and extract the UAttributes.
https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/5/files#diff-5cdcc3261dd77eadb37bcd60c5f85affcc9306454ae7bb8128e6d096ae51a631R61
Maybe we can still keep UAttributesBuilder here?

@sophokles73  Do you have any suggestions?

sophokles73:
My understanding is that you need the UAttributes because you want to put their protobuf encoding into the zenoh message attachment, right? Maybe we can add a
```rust
pub fn request(message_id: UUID, method: UUri, reply_to_address: UUri, options: CallOptions) -> UAttributes
```
factory function to `UAttributes` for this specific purpose? I do not believe that we need a fully flexible builder at this level.

evshary:
Correct. I think this is a proper solution.

PLeVasseur:
I think the proposed function should also take a request ID as well, like this:
```rust
pub fn request(message_id: UUID, method: UUri, request_id: UUID, reply_to_address: UUri, options: CallOptions) -> UAttributes
```
because the UAttributes formed here should be "signed" with a UUID built using a UUIDBuilder from the current uE. So there would be a call to e.g. `uuid_builder.build()` first to get the `request_id` argument to pass in.

sophokles73:
@PLeVasseur based on the current `UMessageBuilder` code you are right. However, I am working on several changes/improvements already. One of them being the insight, that request messages do not carry a separate request_id but instead the message_id is used as the correlation ID when a response wants to reply.

PLeVasseur:
OK. I'll hang tight and wait to check out the PR :slightly_smiling_face:

sophokles73:
closed via #64

PR #59, opened 2024-03-11
Fix/improve UMessageBuilder and UAttributesValidator

sophokles73:
Changed UMessageBuilder to support being used as a message template
for creating multiple message consecutively using different message
ID and payload only.

Moved message type specific checks of UAttributesValidator to
concrete validator structs to improve encapsulation.

### DIFF
--- a/src/uattributes/uattributesvalidator.rs
+++ b/src/uattributes/uattributesvalidator.rs
@@ -15,7 +15,7 @@ use std::time::SystemTime;
 
 use protobuf::Enum;
 
-use crate::{UAttributes, UCode, UMessageType, UriValidator, UUID};
+use crate::{UAttributes, UCode, UMessageType, UPriority, UriValidator, UUID};
 
 use crate::UAttributesError;
 
@@ -161,6 +161,32 @@ pub trait UAttributesValidator {
         }
         Ok(())
     }
+}
+
+/// Verifies that a set of attributes contains a priority that is appropriate for an RPC request message.
+///
+/// # Errors
+///
+/// If [`UAttributes::priority`] contains a value that is less [`UPriority::UPRIORITY_CS4`].
+pub fn validate_rpc_priority(attributes: &UAttributes) -> Result<(), UAttributesError> {
+    attributes
+        .priority
+        .enum_value()
+        .map_err(|unknown_code| {
+            UAttributesError::ValidationError(format!(
+                "RPC message must have a valid priority [{}]",
+                unknown_code
+            ))
+        })
+        .and_then(|prio| {
+            if prio.value() < UPriority::UPRIORITY_CS4.value() {
+                Err(UAttributesError::ValidationError(
+                    "RPC message must have a priority of at least CS4".to_string(),
+                ))
+            } else {
+                Ok(())
+            }
+        })
 }
 
 /// Enum that hold the implementations of uattributesValidator according to type.
@@ -338,6 +364,7 @@ impl UAttributesValidator for RequestValidator {
     /// * [`UAttributesValidator::validate_source`]
     /// * [`UAttributesValidator::validate_sink`]
     /// * [`RequestValidator::validate_permission_level`]
+    /// * `validate_rpc_priority`
     fn validate(&self, attributes: &UAttributes) -> Result<(), UAttributesError> {
         let error_message = vec![
             self.validate_type(attributes),
@@ -346,6 +373,7 @@ impl UAttributesValidator for RequestValidator {
             self.validate_source(attributes),
             self.validate_sink(attributes),
             self.validate_permission_level(attributes),
+            validate_rpc_priority(attributes),
         ]
         .into_iter()
         .filter_map(Result::err)
@@ -471,6 +499,7 @@ impl UAttributesValidator for ResponseValidator {
     /// * [`UAttributesValidator::validate_sink`]
     /// * [`ResponseValidator::validate_reqid`]
     /// * [`ResponseValidator::validate_commstatus`]
+    /// * `validate_rpc_priority`
     fn validate(&self, attributes: &UAttributes) -> Result<(), UAttributesError> {
         let error_message = vec![
             self.validate_type(attributes),
@@ -480,6 +509,7 @@ impl UAttributesValidator for ResponseValidator {
             self.validate_sink(attributes),
             self.validate_reqid(attributes),
             self.validate_commstatus(attributes),
+            validate_rpc_priority(attributes),
         ]
         .into_iter()
         .filter_map(Result::err)
@@ -660,9 +690,9 @@ mod tests {
         }
     }
 
-    #[test_case(Some(UUIDBuilder::new().build()), Some(method_to_invoke()), Some(reply_to_address()), None, Some(2000), None, true; "succeeds for mandatory attributes")]
-    #[test_case(Some(UUIDBuilder::new().build()), Some(method_to_invoke()), Some(reply_to_address()), Some(1), Some(2000), Some(String::from("token")), true; "succeeds for valid attributes")]
-    #[test_case(None, Some(method_to_invoke()), Some(reply_to_address()), Some(1), Some(2000), Some(String::from("token")), false; "fails for missing message ID")]
+    #[test_case(Some(UUIDBuilder::new().build()), Some(method_to_invoke()), Some(reply_to_address()), None, Some(2000), Some(UPriority::UPRIORITY_CS4), None, true; "succeeds for mandatory attributes")]
+    #[test_case(Some(UUIDBuilder::new().build()), Some(method_to_invoke()), Some(reply_to_address()), Some(1), Some(2000), Some(UPriority::UPRIORITY_CS4), Some(String::from("token")), true; "succeeds for valid attributes")]
+    #[test_case(None, Some(method_to_invoke()), Some(reply_to_address()), Some(1), Some(2000), Some(UPriority::UPRIORITY_CS4), Some(String::from("token")), false; "fails for missing message ID")]
     #[test_case(
         Some(UUID {
             // invalid UUID version (not 0b1000 but 0b1010)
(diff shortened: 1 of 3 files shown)


src/umessage/umessagebuilder.rs
```diff
@@ -276,31 +278,87 @@ impl<'a> UMessageBuilder<'a> {
...
+    ///                     .with_message_id(uuid_builder.build())
+    ///                     .build_with_payload("open".into(), UPayloadFormat::UPAYLOAD_FORMAT_TEXT)?;
+    /// assert_ne!(message_one.attributes.id, message_two.attributes.id);
+    /// assert_eq!(message_one.attributes.source, message_two.attributes.source);
+    /// assert_eq!(message_one.attributes.priority, UPriority::UPRIORITY_CS2.into());
+    /// assert_eq!(message_two.attributes.priority, UPriority::UPRIORITY_CS2.into());
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn with_message_id(&mut self, message_id: UUID) -> &mut UMessageBuilder {
```
PLeVasseur, line 353: I really like your proposal to be able to reuse `UMessageBuilder` in client code.

One pattern that I've seen used elsewhere is to have an 'intermediate' builder and then when the builder has been 'finalized' only then allow to call `build()`. Can help prevent building a message that doesn't have what needed.

So, for example, up till we call `with_message_id()`, the type could be `UMessageBuilder` and then after both of those are called and we have `message_id` set, we could return a type `UMessageBuildable` (that'd be temporary, ephemeral).

And then we'd impl `build_with_payload()` only on `UMessageBuildable`.

What do you think? Kind of only gave a rough sketch

src/uattributes/uattributesvalidator.rs
```diff
@@ -360,193 +250,253 @@ impl UAttributesValidators {
...
+    ///
+    /// # Errors
+    ///
+    /// If [`UAttributes::permission_level`] contains a value that is less than 1, an error is returned.
+    pub fn validate_permission_level(
+        &self,
+        attributes: &UAttributes,
+    ) -> Result<(), UAttributesError> {
+        if let Some(plevel) = attributes.permission_level {
+            if plevel < 1 {
```
PLeVasseur, line 309: Isn't 0 okay?

[link to spec](https://github.com/eclipse-uprotocol/up-spec/blob/5f7ea5974b3d74d74c5f013a98cdde575d8708be/basics/qos.adoc#L43)

image

src/uattributes/uattributesvalidator.rs
```diff
@@ -360,193 +250,253 @@ impl UAttributesValidators {
...
-                ))),
-            },
+    /// * [`UAttributesValidator::validate_type`]
+    /// * [`UAttributesValidator::validate_id`]
+    /// * [`UAttributesValidator::validate_ttl`]
+    /// * [`UAttributesValidator::validate_source`]
+    /// * [`UAttributesValidator::validate_sink`]
+    /// * [`RequestValidator::validate_permission_level`]
+    fn validate(&self, attributes: &UAttributes) -> Result<(), UAttributesError> {
+        let error_message = vec![
```
PLeVasseur, line 338: Nice! So we can return multiple different things that "went wrong". :slightly_smiling_face:

src/uattributes/uattributesvalidator.rs
```diff
@@ -360,193 +250,253 @@ impl UAttributesValidators {
...
+    /// Verifies that a set of attributes representing an RPC request contain a valid time-to-live.
     ///
-    /// # Arguments
+    /// # Errors
     ///
-    /// * `attributes` - `UAttributes` object containing the ttl to validate.
+    /// Returns an error if [`UAttributes::ttl`] (time-to-live) is empty or contains a value less than 1.
+    fn validate_ttl(&self, attributes: &UAttributes) -> Result<(), UAttributesError> {
+        match attributes.ttl {
+            Some(ttl) if ttl > 0 => Ok(()),
```
PLeVasseur, line 396: Looks like to me ttl of 0 is still valid, it just means "don't timeout".

[Reference here](https://github.com/eclipse-uprotocol/up-core-api/blob/dc353878914b3459edf2e04531e9cadfc752a2e9/uprotocol/uattributes.proto#L64-L68)

src/uattributes/uattributesvalidator.rs
```diff
@@ -360,193 +250,253 @@ impl UAttributesValidators {
...
+            Some(invalid_ttl) => Err(UAttributesError::validation_error(format!(
```
PLeVasseur, line 397: Maybe some of this handling is a remnant of before Steven's PR merged updating it to a `u32`?

src/uattributes/uattributesvalidator.rs
sophokles73: This check is about the *permission level*, not the *priority* ...

src/uattributes/uattributesvalidator.rs
sophokles73: A request message actually MUST have a TTL > 0

https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/cloudevents.adoc#21-common-attributes

This seems to be inconsistent with the definition of [CallOptions](https://github.com/eclipse-uprotocol/up-core-api/blob/main/uprotocol/uattributes.proto), though (sigh).

src/uattributes/uattributesvalidator.rs
PLeVasseur: My bad... :sweat:

src/uattributes/uattributesvalidator.rs
PLeVasseur: I felt that sigh through my screen hah.

So does that mean we'd need two different behaviors or to reconcile this discrepancy with CallOptions?

src/uattributes/uattributesvalidator.rs
PLeVasseur: Okay, but [over here in the spec](https://github.com/eclipse-uprotocol/up-spec/blob/5f7ea5974b3d74d74c5f013a98cdde575d8708be/basics/permissions.adoc#22-service-declaration-of-upls) it says that it can be 0 and that is the default, right?

image

I brought this up a week or so ago on the PR which was going through that changed this and others to unsigned ints.

src/uattributes/uattributesvalidator.rs
```diff
@@ -27,79 +27,69 @@ use crate::UAttributesError;
...
+    /// Returns an error if [`UAttributes::ttl`] (time-to-live) contains a value greater than 0, but
+    /// * the message has expired according to the timestamp extracted from [`UAttributes::id`] and the time-to-live value, or
+    /// * the current system time cannot be determined.
     fn is_expired(&self, attributes: &UAttributes) -> Result<(), UAttributesError> {
         let ttl = match attributes.ttl {
-            Some(t) if t > 0 => t,
-            Some(_) => return Ok(()),
-            None => 0,
+            Some(t) if t > 0 => {
+                u64::try_from(t).expect("should have been able to convert positive i32 to u64")
```
evshary, line 90: I think this will not be a problem after [the PR](https://github.com/eclipse-uprotocol/up-core-api/commit/1c750092f0621b07e0f41d3ddc7516a0f8c195a9).

src/uattributes/uattributesvalidator.rs
evshary: I also like this idea.

src/umessage/umessagebuilder.rs
```diff
@@ -387,14 +450,16 @@ impl<'a> UMessageBuilder<'a> {
...
+    ///                     .with_message_id(uuid_builder.build())
+    ///                     .with_token(token.clone())
+    ///                     .build_with_payload("lock".into(), UPayloadFormat::UPAYLOAD_FORMAT_TEXT)?;
     /// assert_eq!(message.attributes.token, Some(token));
     /// # Ok(())
     /// # }
     /// ```
-    pub fn with_token(&'a mut self, token: &'a String) -> &'a mut UMessageBuilder {
+    pub fn with_token(&mut self, token: String) -> &mut UMessageBuilder {
+        assert!(self.message_type == UMessageType::UMESSAGE_TYPE_REQUEST);
```
evshary, line 462: I guess the reason you use assert instead of using `Result` is to make the code cleaner, right?

evshary (approved):
LGTM
BTW, maybe I missed it, but do you want to include [factory function to UAttributes](https://github.com/eclipse-uprotocol/up-rust/issues/55#issuecomment-1973013825) in the PR.

src/umessage/umessagebuilder.rs
sophokles73: Returning a `Result` would be different from the value returned by all the other setters and thus prevent fluent API usage. Also, I guess in this case it would be ok to panic because it would be a programming error to include a token in a message other than a request.

sophokles73:
> LGTM BTW, maybe I missed it, but do you want to include [factory function to UAttributes](https://github.com/eclipse-uprotocol/up-rust/issues/55#issuecomment-1973013825) in the PR.

I have a follow-up PR adding the factory function in my local workspace :-)

src/uattributes/uattributesvalidator.rs
sophokles73: FMPOV it makes sense to require a request to always time-out after a certain time. Publish messages, on the other hand, might *live forever*, if so desired ...
So yes, I guess we need two different checks, depending on the type of message.

src/uattributes/uattributesvalidator.rs
sophokles73: Hmm, seems like you are right. I will change it accordingly ...

src/uattributes/uattributesvalidator.rs
PLeVasseur: > FMPOV it makes sense to require a request to always time-out after a certain time. Publish messages, on the other hand, might _live forever_, if so desired ...

Yah, it'd be weird to be like hey here's my request, don't bother getting back to me :sweat_smile: 

In this case should we open an issue on `up-core-api` to clarify that the ttl of 0 should only be valid for `Publish`, `Notification` and `Response`, but not for `Request`?

Better that we align on this I guess at the higher level of spec / core api IMHO.

WDYT?

sophokles73: I am currently preparing a PR for https://github.com/eclipse-uprotocol/up-spec/issues/71 which will include this as well ...

sophokles73:
@PLeVasseur would you mind taking another look? I have added some code for also making sure that requets/response priority is >= CS4 (see https://github.com/eclipse-uprotocol/up-spec/issues/71)

src/uattributes/uattributesvalidator.rs
```diff
@@ -163,6 +163,32 @@ pub trait UAttributesValidator {
...
+        .priority
+        .enum_value()
+        .map_err(|unknown_code| {
+            UAttributesError::ValidationError(format!(
+                "RPC message must have a valid priority [{}]",
+                unknown_code
+            ))
+        })
+        .and_then(|prio| {
+            if prio.value() < UPriority::UPRIORITY_CS4.value() {
```
PLeVasseur, line 182: :+1:

src/uattributes/uattributesvalidator.rs
```diff
@@ -346,6 +373,7 @@ impl UAttributesValidator for RequestValidator {
             self.validate_source(attributes),
             self.validate_sink(attributes),
             self.validate_permission_level(attributes),
+            validate_rpc_priority(attributes),
```
PLeVasseur, line 376: The part of my brain that loves congruity is screaming out in pain right now, but I suppose you don't actually need anything other than `atttributes` for this :)

src/uattributes/uattributesvalidator.rs
PLeVasseur: Just making sure on this one -- did you intend to change this now
```rust
Some(ttl) if ttl > 0 => Ok(()),
```
to align with what we chatted about up-thread or wait on the spec change and do in a follow-up?

src/uattributes/uattributesvalidator.rs
```diff
@@ -804,27 +798,28 @@ mod tests {
         }),
         None,
         None,
-        None,
+        Some(UPriority::UPRIORITY_CS4),
         false;
         "fails for invalid request id")]
-
+    #[allow(clippy::too_many_arguments)]
```
PLeVasseur, line 804: I use this last week lmao

sophokles73:
@PLeVasseur are we good to merge this one?

src/uattributes/uattributesvalidator.rs
sophokles73: I was not aware of any changes that are necessary here? Can you elaborate?

PLeVasseur: D'oh! Missed that this was within the impl for Requests and Publish will use the default impl of `validate_ttl()`, which doesn't put a requirement on ttl to be non-zero.

PLeVasseur (approved):
LGTM! :+1:

PR #60, opened 2024-03-12
`UTransport::registerListener()` using marker trait dispatch for message types + generic

PLeVasseur:
Hey there folks :wave: 

After chatting with @tamarafischer and @stevenhartley and reading through the [equivalent PR over on `up-java`](https://github.com/eclipse-uprotocol/up-java/pull/93), I tried my hand at something similar over here in Rust.

This is very much in draft status to garner feedback! :slightly_smiling_face:

### DIFF
--- a/src/ulistener.rs
+++ b/src/ulistener.rs
@@ -164,12 +164,10 @@ mod tests {
                             return true;
                         }
                     }
-                    Err(_e) => {
-                        return false;
-                    }
+                    Err(_e) => {}
                 }
             }
-            return false;
+            false
         }
     }
 
--- a/src/ulistener.rs
+++ b/src/ulistener.rs
@@ -11,40 +11,46 @@
  * SPDX-License-Identifier: Apache-2.0
  ********************************************************************************/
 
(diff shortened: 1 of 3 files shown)


src/ulistener.rs
```diff
@@ -0,0 +1,223 @@
...
+pub trait NotificationUListener: UListener + Send + Sync {}
+
+pub trait PublishUListener: UListener + Send + Sync {}
+
+pub trait RequestUListener: UListener + Send + Sync {}
+
+pub trait ResponseUListener: UListener + Send + Sync {}
+
+pub trait GenericUListener: UListener + Send + Sync {
+    fn is_match(&self, message: UMessage) -> bool;
```
sophokles73, line 29: Is there any value in having this function? If the transport needs to pass in the message, it can as well simply invoke `onMessage(UMessage)` and let the listener decide if it wants to process or not, or am I mistaken?

PLeVasseur: Ah... that's true I think!

PLeVasseur:
Closing this PR, we're going with another solution

PR #66, opened 2024-03-15
Compatibility with up_core_api 1.5.7

AnotherDaniel:
Adaptions and fixes to compile and test with up-core-api v1.5.7


src/uattributes.rs
```diff
@@ -95,15 +99,14 @@ impl UAttributes {
...
-        let ttl = i32::try_from(options.timeout()).unwrap_or(i32::MAX);
         Self {
             type_: UMessageType::UMESSAGE_TYPE_REQUEST.into(),
             id: Some(message_id).into(),
             priority: UPriority::UPRIORITY_CS4.into(),
             source: Some(reply_to_address).into(),
             sink: Some(method).into(),
-            ttl: Some(ttl),
-            token: options.token().map(|s| s.to_string()),
+            ttl: Some(options.ttl),
```
sophokles73, line 108: can you also update the doc comment accordingly?

src/uattributes/uattributesvalidator.rs
```diff
@@ -86,14 +86,14 @@ pub trait UAttributesValidator {
     /// * the current system time cannot be determined.
     fn is_expired(&self, attributes: &UAttributes) -> Result<(), UAttributesError> {
         let ttl = match attributes.ttl {
-            Some(t) if t > 0 => {
-                // this will not be needed anymore once up-core-api 1.5.7 is released and
-                // ttl is defined as a uint32
-                u64::try_from(t).expect("should have been able to convert positive i32 to u64")
-            }
+            Some(t) => u64::from(t),
```
sophokles73, line 89: IMHO we should change this to
```rust
Some(t) if t > 0 => u64::from(t),
_ => return Ok(()),
```

and then remove the block below
```rust
if ttl == 0 {
  return Ok(());
}
```

src/uattributes/uattributesvalidator.rs
```diff
@@ -115,16 +114,10 @@ pub trait UAttributesValidator {
...
-    /// This default implementation returns an error if [`UAttributes::ttl`] (time-to-live) contains a value less than 0.
-    fn validate_ttl(&self, attributes: &UAttributes) -> Result<(), UAttributesError> {
-        match attributes.ttl {
-            Some(t) if t < 0 => Err(UAttributesError::validation_error(format!(
-                "TTL must not be negative [{t}]"
-            ))),
-            _ => Ok(()),
-        }
+    /// This default implementation always returns Ok(), as u32 ttl can never be < 0
+    fn validate_ttl(&self, _attributes: &UAttributes) -> Result<(), UAttributesError> {
```
sophokles73, line 118: if there is nothing to validate by default, then I would propose to remove the function from the trait and instead add it to the RequestValidator only.

src/cloudevents.rs
```diff
@@ -145,7 +146,10 @@ impl TryFrom<UMessage> for cloudevents::Event {
             event.set_extension("sink", ExtensionValue::String(uri));
         }
         if let Some(commstatus) = attributes.commstatus {
-            event.set_extension("commstatus", ExtensionValue::Integer(commstatus as i64));
+            event.set_extension(
+                "commstatus",
+                ExtensionValue::Integer(commstatus.enum_value_or_default().value() as i64),
```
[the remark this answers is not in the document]
AnotherDaniel: done

src/uattributes.rs
AnotherDaniel: yes... wonder when I'll begin to remember to do this without prompting...

src/uattributes/uattributesvalidator.rs
AnotherDaniel: nice

src/uattributes/uattributesvalidator.rs
AnotherDaniel: done

sophokles73 (approved):
LGTM

Issue #65, opened 2024-03-15
register/unregiter listener: register takes callback function as input but unregister requires string. Other sdk's take callback function as input for both register and unregister listner


sagarsshah:
I wanted to discuss a potential overhead  to our rust SDK's listener apis. Currently, the register_listener() function within the SDK requires a listener parameter of type Box<dyn Fn(Result<UMessage, UStatus>) + Send + Sync + 'static>, returning a listener string for later use with unregister_listener().
However, I've found it challenging to manage listeners effectively without associating them directly with a UUri and associated function. Other SDKs like Java and Python, which accept the listener function directly for both registration and unregistration, our current Rust SDK requires additional bookkeeping to track listeners.
 
Though in real life scenario same uE will rarely listen twice on same UUri but still this can be good correction. Either we expose functions to convert listener function to listener string to uE developer OR we modify rust SDK to match other SDKs.

PLeVasseur:
Hey @sagarsshah :wave:

Thanks for your feedback. I'll make a strawman proposal PR for us to all look at.

PLeVasseur:
Hi @sagarsshah -- check out the draft PR here: #67

PLeVasseur:
I took a different approach over in #68 that is a lot cleaner.

PLeVasseur:
Latest attempt is in #69, using _instance equality_ instead of _type-level equality_ to align with what `up-client-android-java` is doing (and honestly, which makes more sense).

sagarsshah:
Thanks. let me try #69 .

sagarsshah:
I found integration to be very convenient, although I haven't tested it yet. I'll soon share my up-rust-socket code with you for your preview. Thanks.

PLeVasseur:
Hey @sagarsshah -- we still need to get #69 reviewed before I want you to go to far in using it :slightly_smiling_face: 

Let's wait a little longer on others to review the PR

PLeVasseur:
Hey again @sagarsshah -- this was implemented in #69.

Could you close the issue?

sagarsshah:
I've completed the integration with my rust socket transport, and it seems to be ok. However, I still need to conduct testing to ensure everything is working as expected. hopefully by this weekend I will close this issue.

PLeVasseur:
> I've completed the integration with my rust socket transport, and it seems to be ok. However, I still need to conduct testing to ensure everything is working as expected. hopefully by this weekend I will close this issue.

[...]

PR #67, opened 2024-03-15
Aligning with `UTransport::register_listener()` and `UTransport::unregister_listener()` with `UListener`

PLeVasseur:
Hey there :wave: 

Based on #65, I took my shot at aligning our `UTransport` trait closer to the spec.

In particular, I have removed the requirement for the implementation to explicitly manage listeners using `String` identifiers, instead leaning on concrete implementations of the `UListener` trait.

This change is to align with [`registerListener()`](https://github.com/eclipse-uprotocol/up-spec/tree/main/up-l1#22-registerlistener) and [`unregisterListener()`](https://github.com/eclipse-uprotocol/up-spec/tree/main/up-l1#22-registerlistener) by bringing in a [`UListener`](https://github.com/eclipse-uprotocol/up-spec/tree/main/up-l1#1-ulistener) trait.

As you read the code, you can definitely see there were compromises that had to be made to accomplish this.

Let me list the ways:
* `UListener` must have another function `as_any()` which must be implemented by the concrete impls of it to always return `self`
   * IMHO not a _huge_ deal, but we could go a step farther by writing a proc macro that people would call instead which could ensure this is always performed
   * I'm also open to any suggestions on how `UListener` could be tweaked to remove this requirement!
* If we wish to `unregister_listener()`, we must instantiate another `Box<dyn UListener>` from the concrete type to do so
   * Kinda annoying, but not a deal-breaker IMHO

In addition, it does make it quite a bit more annoying because instead of being able to pass in any boxed function anywhere that meets the signature, you must now first implement UListener for each and every different function that you wish to be able to register with `UTransport`. This is not a small change and pretty annoying in my opinion.

I'm looking to see how this would impact current and future implementations of `UTransport` for UPClients. Feedback would be great from @AnotherDaniel, @sophokles73, @evshary in particular :slightly_smiling_face:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,3 +60,4 @@ ureq = { version = "2.9.1", features = ["proxy-from-env"] }
 
 [dev-dependencies]
 test-case = { version = "3.3" }
+async-std = { version = "1.12.0" }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -73,6 +73,7 @@ pub use ustatus::{UCode, UStatus};
 
 mod utransport;
 pub use utransport::UTransport;
+pub mod ulistener;
 
 mod uuid;
 pub use uuid::{UUIDBuilder, UUID};
@@ -97,6 +98,7 @@ pub use up_core_api::utwin;
 // cloudevent-proto, generated and augmented types
 #[cfg(feature = "cloudevents")]
 pub mod cloudevents;
+
 #[cfg(feature = "cloudevents")]
 mod proto_cloudevents {
     include!(concat!(env!("OUT_DIR"), "/cloudevents/mod.rs"));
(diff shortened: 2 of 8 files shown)


AnotherDaniel:
Hm - we explicitly removed the UListener trait approach in favour of a simple listener ID, back in the good old days (round of reviews for the initial PR). 
To be honest, every time I'm beginning to play around with code implementing one of these interfaces, something more fundamental comes along (like up-rust improvements... many to go), and I get detracted again - so myself I've still not really done code that actually uses these interfaces :-(

So on this question, I'd look at @sophokles73 for informed opinions, for now.

PLeVasseur:
Closing in favor of #68 which resolves all of the issues I had with this approach.

PR #68, opened 2024-03-16
Aligning with `UTransport::register_listener()` and `UTransport::unregister_listener()` with `UListener` (this time using generics)

PLeVasseur:
Hey there :wave: 

Based on #65, I took my shot at aligning our `UTransport` trait closer to the spec.

In particular, I have removed the requirement for the implementation to explicitly manage listeners using `String` identifiers, instead leaning on concrete implementations of the `UListener` trait.

This change is to align with [`registerListener()`](https://github.com/eclipse-uprotocol/up-spec/tree/main/up-l1#22-registerlistener) and [`unregisterListener()`](https://github.com/eclipse-uprotocol/up-spec/tree/main/up-l1#22-registerlistener) by bringing in a [`UListener`](https://github.com/eclipse-uprotocol/up-spec/tree/main/up-l1#1-ulistener) trait.

As you read the code, you can see that there was a need for a new wrapper type called `ListenerWrapper` around UListener in order to allow us to disambiguate between concrete impls of it. That `ListenerWrapper` need only be used in the `up-client-foo-rust` implementation. uEs just impl `UListener`.

Largely resolves the issues I had with the first attempt over in #67 thanks to that wrapper type.

I'm looking to see how this would impact current and future implementations of `UTransport` for UPClients. Feedback would be great from @AnotherDaniel, @sophokles73, @evshary in particular :slightly_smiling_face:


PLeVasseur:
Closing in favor of #69, where we use the same semantics as `up-client-android-java` of using _reference equality_ of UListener instances instead of _type equality_ as we were in this approach.

Issue #58, opened 2024-03-11
UUIDBuilder -- should it be a singleton as it is in the other language libs? (e.g. `up-java` and `up-cpp`)


PLeVasseur:
Howdy :wave: 

@stevenhartley recently [brought up](https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/12#pullrequestreview-1923427766) on a PR going into `up-client-zenoh-rust` that the other language libs implement their UUID builder / factory as singletons to avoid an end-user accidentally foot-gunning themselves by using multiple UUIDBuilder within one uE.

He asked why the same wasn't done in `up-rust`. I had my thoughts as to why:

Making it a singleton in Rust forces upon the end user a certain way of interacting with the singleton and there are multiple (reasonable) ways of doing this in Rust. However, maybe it makes sense to give a "simple" path to an end-user not foot-gunning themselves with multiple UUIDBuilder, but allow the possibility for an advanced user to ensure it's a singleton themselves?

A couple of reasonable ways of doing this in Rust:
* lazy_static
* once_cell

Looking to hear feedback on this, hopefully from @sophokles73, @AnotherDaniel, and @evshary

evshary:
I'm fine with putting singleton in either `up-rust` or `up-client-zenoh-rust`.
I'm wondering if there is any special use case we need to have multiple UUIDBuilder.

PLeVasseur:
Bump =^)

Hoping to hear feedback from @sophokles73 and @AnotherDaniel in particular

sophokles73:
Sorry for the delay.

If I understand correctly, the main point here is that we want to be able to share a *single instance* of UUIDBuilder within a uEntity for creating UUIDs, right?
If so, then I guess we should focus on making UUIDBuilder thread-safe first. After that we can discuss whether we want to use the Singleton pattern for sharing the instance or pass around a reference explicitly.

PLeVasseur:
> If I understand correctly, the main point here is that we want to be able to share a _single instance_ of `UUIDBuilder` within a uEntity for creating UUIDs, right?

Yup, ideally to prevent the foot-gun of end-users who are developing uEs doing the "wrong" thing by e.g. creating a new `UUIDBuilder` and building a UUID from it on e.g. every publish.

> If so, then I guess we should focus on making UUIDBuilder thread-safe first. After that we can discuss whether we want to use the Singleton pattern for sharing the instance or pass around a reference explicitly.

Makes sense.

------------------

[...]

PR #74, opened 2024-03-29
Guarantee uE implementors always get the correct UUID when calling `UUIDBuilder::build()` by using internally held singleton

PLeVasseur:
Addresses #58 

## What

Updated `UUIDBuilder` to expose only `build()` to allow end-users a simpler, more consistent API when generating UUIDs for their uE.

## Why

End users of the library should be shielded from making mistakes, which is what we do here. They simply call `UUIDBuilder::build()` and they'll always get the correct result.

## How

We use an internally held instance of `UUIDBuilder` which generates the `rand_b` portion once, meaning all calls to `build()` have the same `rand_b` in the UUIDs.

We use a compare-and-swap (CAS) technique using an AtomicU64 to ensure we correctly handled logic for generating timestamp + counter portion, even in extremely high-contention scenarios when generating UUIDs. To do so we had to refactor so that we regenerate the timestamp at the top of the CAS loop, rather than be passed in the timestamp to use.

## Mutex vs Atomic Compare-and-Swap (CAS) Performance

Using the Atomic CAS technique is approximately twice as fast as using a Mutex

With 10 async tasks, each generating 400 UUIDs all released and run at once
 * Mutex: ~2ms
 * Atomic CAS: ~0.8ms


src/uattributes/uattributesvalidator.rs
```diff
@@ -590,17 +593,22 @@ mod tests {
             ttl,
             ..Default::default()
         };
+
+        if let Some(created_millis_ago) = created_millis_ago {
+            thread::sleep(std::time::Duration::from_millis(created_millis_ago));
```
evshary, line 598: This is more intuitive to create a timeout UAttributes, but I'm not sure whether this will need a longer time to do the test. (Maybe it's ok while we don't have too many of this kind of tests).

src/uattributes/uattributesvalidator.rs
```diff
@@ -556,29 +558,30 @@ mod tests {
...
+    #[test_case(UMessageType::UMESSAGE_TYPE_RESPONSE, None, None, None, false; "for Response message without ID nor TTL")]
+    #[test_case(UMessageType::UMESSAGE_TYPE_RESPONSE, None, None, Some(0), false; "for Response message without ID with TTL 0")]
+    #[test_case(UMessageType::UMESSAGE_TYPE_RESPONSE, None, None, Some(500), false; "for Response message without ID with TTL")]
+    #[test_case(UMessageType::UMESSAGE_TYPE_RESPONSE, Some(1000), Some(UUIDBuilder::build()), None, false; "for Response message with ID without TTL")]
+    #[test_case(UMessageType::UMESSAGE_TYPE_RESPONSE, Some(1000), Some(UUIDBuilder::build()), Some(0), false; "for Response message with ID and TTL 0")]
+    #[test_case(UMessageType::UMESSAGE_TYPE_RESPONSE, Some(1000), Some(UUIDBuilder::build()), Some(500), true; "for Response message with ID and expired TTL")]
+    #[test_case(UMessageType::UMESSAGE_TYPE_RESPONSE, Some(1000), Some(UUIDBuilder::build()), Some(2000), false; "for Response message with ID and non-expired TTL")]
     fn test_is_expired(
         message_type: UMessageType,
+        created_millis_ago: Option<u64>,
```
evshary, line 584: We don't need Option here and only sleep while the created_millis_ago is not 0.

src/uuid/uuidbuilder.rs
```diff
@@ -33,174 +32,196 @@ pub struct UUIDBuilder {
...
     ///
-    /// let builder = UUIDBuilder::default();
-    /// let uuid = builder.build();
-    /// assert!(uuid.is_uprotocol_uuid());
-    /// ```
-    fn default() -> Self {
-        Self::new()
+    /// UUID with consistent `rand_b` portion, which uniquely identifies this uE
+    pub fn build() -> UUID {
+        #[allow(clippy::redundant_closure)]
```
evshary, line 43: Maybe we can have this:
`static UUIDBUILDER_SINGLETON: Lazy<UUIDBuilder> = Lazy::new(UUIDBuilder::new);`

src/uuid/uuidbuilder.rs
```diff
@@ -33,174 +32,196 @@ pub struct UUIDBuilder {
...
     /// Creates a new UUID for a given timestamp.
     ///
+    /// # Note
+    ///
+    /// We use a compare-and-swap (CAS) technique to attempt to repeatedly build a fresh UUID
+    /// until we succeed. Benchmarks roughly twice as fast as using a Mutex.
+    ///
     /// # Arguments
     ///
     /// * `timestamp` - The timestamp (in milliseconds since UNIX EPOCH) to use.
```
evshary, line 81: Legacy comments?

src/uuid/uuidbuilder.rs
```diff
@@ -33,174 +32,196 @@ pub struct UUIDBuilder {
...
-    /// ```
-    pub fn build(&self) -> UUID {
-        let now = SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .expect("current system time is set to a point in time before UNIX Epoch");
-        let now_millis = u64::try_from(now.as_millis())
-            .expect("current system time is set to a point in time too far in the future");
-        self.build_with_instant(now_millis)
+    /// For internal testing purposes only. For end-users, please use [`UUIDBuilder::build()`]
+    pub(crate) fn build_from_instance(&self) -> UUID {
```
evshary, line 68: How about remove the `build_internal` and only leave `build_from_instance`? Or do you plan to add something in `build_from_instance` in the future?

src/uuid/uuidbuilder.rs
```diff
@@ -33,174 +32,196 @@ pub struct UUIDBuilder {
...
             }
 
-            // set UUID's version to 'custom'
-            self.msb.load(Ordering::SeqCst) & BITMASK_CLEAR_VERSION | crate::uuid::VERSION_CUSTOM
-        };
-
-        UUID::from_u64_pair(new_msb, self.lsb)
-            .expect("should have been able to create UUID for valid timestamp")
+            // Try to update the MSB atomically.
+            match self.msb.compare_exchange(
```
evshary, line 119: Okay, so we use `compare_exchange` instead of `swap` here to ensure no one modifies the `self.msb`, right?
Brilliant idea!

src/uuid/uuidbuilder.rs
```diff
@@ -33,174 +32,196 @@ pub struct UUIDBuilder {
...
+                current_msb,
+                new_msb,
+                Ordering::SeqCst,
+                Ordering::SeqCst,
+            ) {
+                Ok(_) => {
```
evshary, line 125: How about
```rust
// Only return while CAS succeeds
if Ok(_) = self.msb.compare_exchange(current_msb, new_msb, Ordering::SeqCst, Ordering::SeqCst) {
    return UUID::from_u64_pair(new_msb, self.lsb)
                        .expect("should have been able to create UUID for valid timestamp")
}
```

src/uuid/uuidbuilder.rs
```diff
@@ -33,174 +32,196 @@ pub struct UUIDBuilder {
...
+        let end = Instant::now();
 
-        // add 1 millisecond to the maximum duration, to overflow
-        let overflowed_48_bit_unix_ts_ms = max_48_bit_unix_ts_ms + 1;
+        let duration = end.duration_since(start);
 
-        let builder = UUIDBuilder::new();
-        let _uprotocol_uuid_past_max_unix_ts_ms =
-            builder.build_with_instant(overflowed_48_bit_unix_ts_ms);
+        println!("All tasks completed in {:?}.", duration);
```
evshary, line 189: I prefer to use `log` instead of `println`

src/uuid/uuidbuilder.rs
```diff
@@ -33,174 +32,196 @@ pub struct UUIDBuilder {
...
+                }
+            }
+        }
+
+        for (task_id, uuid) in &duplicates {
+            let msb = uuid.msb;
+            let timestamp = msb >> 16;
+            let counter = msb & MAX_COUNT;
+
+            println!("task_id: {task_id} timestamp: {timestamp} counter: {counter}");
```
evshary, line 207: I prefer to use `log` instead of `println`

evshary (commented):
@PLeVasseur I like your idea of using CAS. I left some nitpicks.

src/uattributes/uattributesvalidator.rs
PLeVasseur: Good point!

src/uuid/uuidbuilder.rs
PLeVasseur: Ahhh okay. Makes sense

src/uuid/uuidbuilder.rs
PLeVasseur: I think this is reasonable. :+1:  I'm already doing a superset of the one remaining test.

src/uuid/uuidbuilder.rs
PLeVasseur: Good idea :slightly_smiling_face:

src/uuid/uuidbuilder.rs
PLeVasseur: Yup, should swap in `log` :+1:

src/uuid/uuidbuilder.rs
```diff
@@ -33,174 +32,171 @@ pub struct UUIDBuilder {
...
-    #[test]
-    fn test_uuid_for_constant_random() {
-        let factory = UUIDBuilder::new();
-        let uuid1 = factory.build();
-        let uuid2 = factory.build();
-        assert_eq!(uuid1.lsb, uuid2.lsb);
-    }
+        let duration = end.duration_since(start);
+
+        info!("All tasks completed in {:?}.", duration);
```
AnotherDaniel, line 164: Is it really worth it/necessary to add the log dependency? I'm not against using proper logging, but does it add sufficient value in this case?

src/uuid/uuidbuilder.rs
```diff
@@ -33,174 +32,171 @@ pub struct UUIDBuilder {
...
+        }
 
-        // add 1 millisecond to the maximum duration, to overflow
-        let overflowed_48_bit_unix_ts_ms = max_48_bit_unix_ts_ms + 1;
+        for (task_id, uuid) in &duplicates {
+            let msb = uuid.msb;
+            let timestamp = msb >> 16;
+            let counter = msb & MAX_COUNT;
+
+            error!("task_id: {task_id} timestamp: {timestamp} counter: {counter}");
```
AnotherDaniel, line 182: Same comment as above

Cargo.toml
```diff
@@ -44,7 +44,10 @@ async-trait = { version = "0.1" }
 bytes = { version = "1.5" }
 chrono = { version = "0.4.32" }
 cloudevents-sdk = { version = "0.7", optional = true }
+env_logger = { version = "0.11.3" }
+log = { version = "0.4" }
```
AnotherDaniel, line 48: I might have missed something - but in the PR, the logger is only used within uuidbuilder tests? Shouldn't this be a dev-dependency then? Also, do we really need to pull this in, compared to what we're actually using it for? 

If we have to do logging, nothing at all against doing it properly. But so far we always got away without...

AnotherDaniel (commented):
Looks good overall, like what you can do with Rust to support the desired functionality, and that you put in the time to figure it out :)

src/uuid/uuidbuilder.rs
PLeVasseur: I'm okay making it a dev dependency as it's only used in tests 👍

I suppose there's the one tricky case of if we generated more than `counter` number of UUIDs per ms that we could put a warn!() for. But realistically this likely won't occur.

What do you think?

AnotherDaniel: Well this we will have to approach in a more generic way, should this be a valid scenario - logging this info during tests will quite surely be lost on the world ;-)

Famous last words would be "we'll never need more than `counter` ids per ms!". To put this differently: do we see ourselves even approaching that boundary in the coming years? Or differently again: if we're afeared of this scenario, we might want to increase the size of counter generally, or ensure in the code that if counter is "full", we don't generate any more ids until the next ms turns around.

I'd favour removing the logging dependency and those two output lines entirely, and rather replace with asserts in some form. E.g., 'assert that we're not overrunning counter within the same ms'.

Does that make sense?

PLeVasseur: These seem pointed at @stevenhartley :slightly_smiling_face: 

> Famous last words would be "we'll never need more than counter ids per ms!". To put this differently: do we see ourselves even approaching that boundary in the coming years? Or differently again: if we're afeared of this scenario, we might want to increase the size of counter generally, or ensure in the code that if counter is "full", we don't generate any more ids until the next ms turns around.

----------

> I'd favour removing the logging dependency and those two output lines entirely, and rather replace with asserts in some form. E.g., 'assert that we're not overrunning counter within the same ms'.
> 
> Does that make sense?

Yup, I can look into this.

src/uuid/uuidbuilder.rs
```diff
@@ -33,174 +32,171 @@ pub struct UUIDBuilder {
...
 
-impl UUIDBuilder {
     /// Creates a new builder for creating uProtocol UUIDs.
     ///
-    /// The same bulder instance can be used to create one or more UUIDs
-    /// by means of invoking [`UUIDBuilder::build`].
-    ///
-    /// # Examples
+    /// The same builder instance can be used to create one or more UUIDs
+    /// by means of invoking [`UUIDBuilder::build_from_instance`].
```
sophokles73, line 50: invoking `UUIDBuilder::build`?

src/uuid/uuidbuilder.rs
```diff
@@ -33,174 +32,171 @@ pub struct UUIDBuilder {
...
-            .expect("current system time is set to a point in time before UNIX Epoch");
-        let now_millis = u64::try_from(now.as_millis())
-            .expect("current system time is set to a point in time too far in the future");
-        self.build_with_instant(now_millis)
-    }
-
     /// Creates a new UUID for a given timestamp.
     ///
-    /// # Arguments
+    /// # Note
```
sophokles73, line 64: the first sentence will need to be adapted: no given timestamp (anymore) ...

src/uuid/uuidbuilder.rs
```diff
@@ -33,174 +32,171 @@ pub struct UUIDBuilder {
...
     ///
-    /// * `timestamp` - The timestamp (in milliseconds since UNIX EPOCH) to use.
+    /// We use a compare-and-swap (CAS) technique to attempt to repeatedly build a fresh UUID
+    /// until we succeed. Benchmarks roughly twice as fast as using a Mutex.
```
sophokles73, line 67: is this information that is relevant to a user? If not, we should move this into a comment in the code.

src/uuid/uuidbuilder.rs
PLeVasseur: > I'd favour removing the logging dependency and those two output lines entirely, and rather replace with asserts in some form. E.g., 'assert that we're not overrunning counter within the same ms'.
> 
> Does that make sense?

Pushed up a change which:
1. Moves the logging deps over to dev dependencies
   * I think these logging lines are still valuable for the case if someone ever changes the test case to have > 10 tasks or > 409 UUIDs to be generated per task to clarify why it broke
2. Clarifies the asserts check if we've overrun the counter

Hope that resolves the issue, if not lemme know :slightly_smiling_face:

src/uattributes/uattributesvalidator.rs
```diff
@@ -590,17 +593,20 @@ mod tests {
             ttl,
             ..Default::default()
         };
+
+        thread::sleep(std::time::Duration::from_millis(created_millis_ago));
+
```
sophokles73, line 598: FMPOV it is an anti-pattern to rely on sleeping in test code. Please replace this with the original or a clock-tick based approach.

src/uuid/uuidbuilder.rs
```diff
@@ -33,174 +32,174 @@ pub struct UUIDBuilder {
...
+                let current_counter = current_msb & MAX_COUNT;
+                if current_counter < MAX_COUNT {
+                    // Prepare new msb with incremented counter.
+                    new_msb = current_msb + 1;
                 } else {
                     // this should never happen in practice because we
                     // do not expect any uEntity to emit more than
                     // 4095 messages/ms
                     // so we simply keep the current counter at MAX_COUNT
+                    new_msb = current_msb
```
sophokles73, line 93: while it might be true that we do not reach this limit easily, I tend to agree with @AnotherDaniel that we might simply want to wait for the new millenium, eh millisecond, here, instead of issuing the same UUID multiple times.

I guess simply `continue`ing should do the trick here, shouldn't it? The price to pay would be a very slightly delayed sending of messages but at least, we would always have unique UUIDs ...

src/uuid/uuidbuilder.rs
sophokles73: With the change I have proposed above, this should never happen again ....

sophokles73 (changes requested):

src/uattributes/uattributesvalidator.rs
PLeVasseur: OK -- I think I was able to pull this off. Can you take another look @sophokles73?

src/uattributes/uattributesvalidator.rs
```diff
@@ -931,13 +932,16 @@ mod tests {
...
     }
 
-    fn create_id_for_timestamp(created_millis_ago: u64) -> UUID {
-        let now = SystemTime::now();
-        let since_epoch_start = now.duration_since(SystemTime::UNIX_EPOCH).unwrap();
-        let now_millis_since_epoch_start = u64::try_from(since_epoch_start.as_millis()).unwrap();
-
-        let creation_time = now_millis_since_epoch_start - created_millis_ago;
-        UUIDBuilder::new().build_with_instant(creation_time)
+    fn uuid_n_ms_in_past(n_ms_in_past: u64) -> UUID {
```
sophokles73, line 935: Maybe we can add this as `pub(crate) fn UUIDBuilder::build_with_instant(timestap) -> UUID` so that it can (potentially) be reused in other test cases of the up-rust crate?

src/uuid/uuidbuilder.rs
```diff
@@ -11,18 +11,17 @@
...
+use rand::random;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use crate::UUID;
 
-const BITMASK_CLEAR_VERSION: u64 = 0xffffffffffff0fff;
-const BITMASK_CLEAR_VARIANT: u64 = 0x3fffffffffffffff;
+pub(crate) const BITMASK_CLEAR_VERSION: u64 = 0xffff_ffff_ffff_0fff;
+const BITMASK_CLEAR_VARIANT: u64 = 0x3fff_ffff_ffff_ffff;
```
sophokles73, line 22: nice formatting :+1:

src/uuid/uuidbuilder.rs
```diff
@@ -33,174 +32,148 @@ pub struct UUIDBuilder {
...
+                let current_counter = current_msb & MAX_COUNT;
+                if current_counter < MAX_COUNT {
+                    // Prepare new msb with incremented counter.
+                    new_msb = current_msb + 1;
                 } else {
                     // this should never happen in practice because we
                     // do not expect any uEntity to emit more than
                     // 4095 messages/ms
                     // so we simply keep the current counter at MAX_COUNT
+                    continue;
```
sophokles73, line 89: maybe extend the comment
> and wait for the next millisecond to arrive.

src/uattributes/uattributesvalidator.rs
PLeVasseur: I could see adding a `pub(crate)` function which does what I wrote here -- generates a UUID and then subtracts a certain amount of time from it.

I ran into issues with passing in an absolute timestamp. Read on to hear of my woes :point_down: 

--------

I was running into a fair amount of tricky debugging and headscratching before I moved away from the two-step process that was here of `build()` -> `build_with_timestamp()`, to calculating the timestamp purely in the loop of the compare-and-swap and not passing in the timestamp.

Imagine on thread 1 we had entered `build_internal()` from `build_with_timestamp()` with a timestamp of n and on thread 2 we had entered `build_internal()` from `build_with_timestamp() with a timestamp of n+1, and the previous time in the `msb` was n.

We make it to thread 2 first, see the `msb` time of n which differs from the passed in timestamp of n+1 and set the timestamp to n+1.

We hit thread 1 second, see the `msb` time of n+1 which differs from the passed in timestamp of n and set the timestamp to n (going back in time!).

If I missed the point though, please feel free to let me know your thoughts :slightly_smiling_face:

sophokles73: > I could see adding a pub(crate) function which does what I wrote here -- generates a UUID and then subtracts a certain amount of time from it.

That is what I had in mind :-) Just didn't want to bury it in the test cases ...

PLeVasseur: :laughing: Can you tell from my paragraph that I had gone through some trauma...

Sounds good, I'll move this over into `UUIDBuilder` so it can be accessed from elsewhere for other tests :+1:

PLeVasseur: Okey dokey -- can you take another look?

sophokles73 (approved):
LGTM

Thanks for contributing :+1:

PR #76, opened 2024-04-04
Make UAttributesValidator Send

sophokles73:
The UAttributesValidator trait is being used by UMessageBuilder for checking that the builder's state is safe before actually creating a message.

Making UAttributesValidator Send will allow UMessageBuilder to be sent between threads, which is useful for creating multi-threaded UTransport implementations.


sophokles73:
@PLeVasseur would you mind taking a look?

PLeVasseur (approved):
Seems reasonable to me.

sophokles73:
I do not want to bother other (mostly unaffected) committers with this trivial change, thus merging this without further review.

PR #79, opened 2024-04-05
Make Utransport-related traits async-usable

AnotherDaniel:
Based on up-streamer review discussion - we want these traits to be send-able between threads.


sophokles73 (approved):
LGTM

PR #81, opened 2024-04-08
Improve versatility of ComparableListener

sophokles73:
Implementing UTransport based on ComparableListener may require
exposing the (interal) ComparableListener type to clients of UTransport
when it comes to invoking the listener.

Added an `into_inner` function for getting a clone of the reference to
the wrapped listener. This way, client code can easily transform a
ComparableListener back to its underlying listener type.

### DIFF
--- a/src/utransport.rs
+++ b/src/utransport.rs
@@ -291,6 +291,10 @@ impl ComparableListener {
     pub fn new(listener: Arc<dyn UListener>) -> Self {
         Self { listener }
     }
+    /// Gets a clone of the wrapped reference to the listener.
+    pub fn into_inner(&self) -> Arc<dyn UListener> {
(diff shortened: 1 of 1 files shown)


sophokles73:
@PLeVasseur would you mind taking a look?

PLeVasseur:
Hi @sophokles73  -- could you help me understand when a `UTransport` impl would need `to_inner()`?

`Deref` is impled on `ComparableListener` to allow getting access to the two functions within the held `dyn UListener`. Do you see a use case beyond that?

Thinking on it for a minute here -- are you interested in being able to pass around the held `Arc<dyn UListener>` instead of a `ComparableListener`?

PLeVasseur (approved):
LGTM

PR #83, opened 2024-04-09
Adding `USubscription` trait

PLeVasseur:
# Why

@stevenhartley realized that it's important that `USubscription` be a trait to allow for its usage as a trait in `up-streamer-rust`

See https://github.com/eclipse-uprotocol/up-java/pull/85/commits/6e19af5cb8ffca7ffee205ef2638ead44cf73d55 for some further details

# What

* Trait methods are documented
* Trait is documented with prose description and doc examples

Tagging in @AnotherDaniel & @sophokles73 for their thoughts

### DIFF
--- a/src/core/usubscription.rs
+++ b/src/core/usubscription.rs
@@ -1,0 +1,308 @@
+pub use crate::up_core_api::usubscription::{
+    subscription_status::State, EventDeliveryConfig, FetchSubscribersRequest,
+    FetchSubscriptionsRequest, FetchSubscriptionsResponse, NotificationsRequest,
+    SubscribeAttributes, SubscriberInfo, SubscriptionRequest, SubscriptionResponse,
+    SubscriptionStatus, UnsubscribeRequest,
+};
+use crate::UStatus;
+use async_trait::async_trait;
+
+/// `USubscription` is the uP-L3 client interface to the uSubscription service.
+///
+/// A client would use a concrete implementation of `USubscription` typically to subscribe to
+/// a topic of interest and then unsubscribe when finished.
+///
+/// Implementations of `USubscription` can be transport-specific to allow for flexibility and optimizations.
+///
+/// # Examples
+///
+/// ## Typical usage
+///
+/// ```
+/// # use async_trait::async_trait;
+/// # use std::future::Future;
+/// # use up_rust::{UMessage, Number, UAuthority, UEntity, UResource, UStatus, UUri, UListener};
+/// #
+/// # mod up_client_foo {
+/// #     use std::sync::Arc;
+/// #     use up_rust::{UTransport, UListener, UStatus, UMessage, UUri};
+/// #     use async_trait::async_trait;
+/// #     pub struct UPClientFoo;
+/// #
+/// #     #[async_trait]
+/// #     impl UTransport for UPClientFoo {
+/// #         async fn send(&self, message: UMessage) -> Result<(), UStatus> {
+/// #             todo!()
+/// #         }
+/// #
+/// #         async fn receive(&self, topic: UUri) -> Result<UMessage, UStatus> {
+/// #             todo!()
+/// #         }
+/// #
+/// #         async fn register_listener(&self, topic: UUri, listener: Arc<dyn UListener>) -> Result<(), UStatus> {
+/// #             Ok(())
+/// #         }
+/// #
+/// #         async fn unregister_listener(&self, topic: UUri, listener: Arc<dyn UListener>) -> Result<(), UStatus> {
+/// #             Ok(())
+/// #         }
+/// #     }
+/// #
+/// #     impl UPClientFoo {
+/// #         pub fn new() -> Self {
+/// #             Self
+/// #         }
+/// #     }
+/// # }
+/// #
+/// # mod usubscription_foo {
+/// #     use async_trait::async_trait;
+/// #     use protobuf::EnumOrUnknown;
+/// #     use up_rust::{UStatus, UCode,
+/// #         core::usubscription::{USubscription, FetchSubscribersRequest, FetchSubscriptionsRequest,
+/// #                               FetchSubscriptionsResponse, NotificationsRequest, SubscriptionRequest,
+/// #                               SubscriptionResponse, UnsubscribeRequest, SubscriptionStatus, State,
+/// #                               EventDeliveryConfig},
+/// #     };
+/// #
+/// #     pub struct USubscriptionFoo;
+/// #
+/// #     #[async_trait]
+/// #     impl USubscription for USubscriptionFoo {
+/// #         async fn subscribe(&self, subscription_request: SubscriptionRequest) -> Result<SubscriptionResponse, UStatus> {
+/// #             let subscription_status = SubscriptionStatus {
+/// #                 state: EnumOrUnknown::from(State::SUBSCRIBED),
+/// #                 code: EnumOrUnknown::from(UCode::OK),
+/// #                 message: "Subscription success".to_string(),
+/// #                 ..Default::default()
+/// #             };
+/// #
+/// #             let event_delivery_config = EventDeliveryConfig {
+/// #                 id: "SUBSCRIPTION_TOPIC".to_string(),
+/// #                 type_: "Foo/Vehicle/EventHubs".to_string(),
+/// #                 attributes: Default::default(),
+/// #                 ..Default::default()
+/// #             };
+/// #
+/// #             let subscription_response = SubscriptionResponse {
+/// #                 status: Some(subscription_status).into(),
+/// #                 config: Default::default(),
+/// #                 topic: Default::default(),
+/// #                 ..Default::default()
+/// #             };
+/// #
+/// #             Ok(subscription_response)
+/// #         }
+/// #
+/// #         async fn unsubscribe(&self, unsubscribe_request: UnsubscribeRequest) -> Result<(), UStatus> {
+/// #             Ok(())
+/// #         }
+/// #
+/// #         async fn fetch_subscriptions(&self, fetch_subscriptions_request: FetchSubscriptionsRequest) -> Result<FetchSubscriptionsResponse, UStatus> {
+/// #             todo!()
+/// #         }
+/// #
+/// #         async fn register_for_notifications(&self, notifications_request: NotificationsRequest) -> Result<(), UStatus> {
+/// #             todo!()
+/// #         }
+/// #
+/// #         async fn unregister_for_notifications(&self, notifications_request: NotificationsRequest) -> Result<(), UStatus> {
+/// #             todo!()
+/// #         }
+/// #
+/// #         async fn fetch_subscribers(&self, fetch_subscribers_request: FetchSubscribersRequest) -> Result<FetchSubscriptionsResponse, UStatus> {
+/// #             todo!()
+/// #         }
+/// #     }
+/// #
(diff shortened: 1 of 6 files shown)


AnotherDaniel (commented):
LGTM to start with. 
We are beginning to collect doc-examples of a complexity and length that make we wonder whether some of these wouldn't be better of as dedicated examples, in a ./examples folder for instance. Not necessarily a must-do for this specific PR, but WDYT in general?

PLeVasseur:
> We are beginning to collect doc-examples of a complexity and length that make we wonder whether some of these wouldn't be better of as dedicated examples, in a ./examples folder for instance. Not necessarily a must-do for this specific PR, but WDYT in general?

Over in the PR for `up-streamer-rust` I saw a lot of code duplication in integration tests I was writing, so factored out setup / helper code.

I could see doing something similar for doc examples where there's lengthy setup that's got #'s in front of them so they don't show up in the generated example code: go look for duplicates and factor them out into a separate non-pub module.

WDYT?

AnotherDaniel:
Yeah, that's one thing that might help. The other thing that irritates me a bit (probably a matter of taste) is that we have 3 pages of docstring for a 3-line function - you know what I mean.
Not saying that this always a bad thing, but that's where I'm coming from with that comment... 

I guess for now we do what you propose above to make the doc a bit more digestible - but I suggest we soon look into extracting the long docs out into an examples collection.

src/lib.rs
```diff
@@ -86,11 +86,18 @@ mod up_core_api {
...
 #[cfg(feature = "udiscovery")]
 pub use up_core_api::udiscovery;
-#[cfg(feature = "usubscription")]
-pub use up_core_api::usubscription;
 #[cfg(feature = "utwin")]
 pub use up_core_api::utwin;
 
+mod core;
+#[cfg(feature = "usubscription")]
+pub use core::{
```
sophokles73, line 94: I would rather not re-export all of these type but keep them in module `core::usubscription`. Anticipating that we will be doing a similar thing next for `utwin` and `udsicovery`, I think we should keep at least some structure.

sophokles73 (changes requested):
I agree with @AnotherDaniel: the doc comment for the USubscription trait should only contain a short explanation and references to the spec. All example implementation code should go into a dedicated module in the `examples` folder ...

sophokles73:
> I wrote that example for the 99% of users of USubscription which will only ever need to call subscribe() and unsubscribe(). IMHO it makes sense to cover this usage here for the non-power user of up-rust.

Those pour souls are already doomed, being forced to use this API ;-)
I have opened https://github.com/eclipse-uprotocol/up-spec/issues/112 to eventually get this right ...

sophokles73:
> Ah okay. Thought it'd be good to have an example available usage for when folks review the documentation of the most common case.
> 
> I wrote that example for the 99% of users of `USubscription` which will only ever need to call `subscribe()` and `unsubscribe()`. IMHO it makes sense to cover this usage here for the non-power user of `up-rust`.
> 
> Other more exotic uses of `USubscription` I agree should be relegated to an examples folder.
> 
> WDYT?

If we want to improve the situation for these people, then we should improve the information in up-spec where this API is defined.

PLeVasseur:
> If we want to improve the situation for these people, then we should improve the information in up-spec where this API is defined.

Sure, I'm definitely on board with improving `up-spec`.

---------

I suppose I'm thinking of the end user that was just told by their team lead that they have to write a uE and hasn't used uProtocol before and just needs to get some basic stuff done to ship on a deadline.

In my opinion and experience, most users will not go review spec documents and we can prevent most issues with the 99% use case of simple subscribe and unsubscribe by putting an example here where the user is most likely to look first.

Remember @sophokles73 that you, me, and @AnotherDaniel are the _outliers_ as people that are reading the spec :sweat_smile: 

(I updated the example to hide / shorten it btw)

WDYT?

src/lib.rs
```diff
@@ -86,11 +86,11 @@ mod up_core_api {
 // Types not used by up_rust, but re-exported to up_rust users, keeping them in their respective submodules
 #[cfg(feature = "udiscovery")]
 pub use up_core_api::udiscovery;
-#[cfg(feature = "usubscription")]
-pub use up_core_api::usubscription;
 #[cfg(feature = "utwin")]
 pub use up_core_api::utwin;
 
+pub mod core;
```
sophokles73, line 92: maybe we can apply the same pattern to `utwin` and `udiscovery` as well? Otherwise we end up with having the `usubscription` types available in `up_rust::core::subscription` only while we have the types of `utwin` and `udiscovery` available in `up_core` ...

PLeVasseur: Yup, makes sense to me. I'll go move them in a similar manner :+1:

sophokles73 (approved):
LGTM

src/core/usubscription.rs
```diff
@@ -0,0 +1,308 @@
...
+    /// Subscribe to a topic, using a [`SubscriptionRequest`]
+    ///
+    /// # Parameters
+    ///
+    /// * `subscription_request` - A request to subscribe
+    ///
+    /// # Returns
+    ///
+    /// * [`SubscriptionResponse`] detailing if subscription was successful with other metadata
+    async fn subscribe(&self, subscription_request: SubscriptionRequest) -> SubscriptionResponse;
```
sophokles73, line 238:
```suggestion
    async fn subscribe(&self, subscription_request: SubscriptionRequest) -> Result<SubscriptionResponse, UStatus>;
```

Can we at least provide for the means to do proper response handling? Even if the service implementation would return a UStatus in the response, the local stub could still map it to the proper Result ...

src/core/usubscription.rs
```diff
@@ -0,0 +1,308 @@
...
+    /// Unsubscribe to a topic, using an [`UnsubscribeRequest`]
+    ///
+    /// # Parameters
+    ///
+    /// * `unsubscribe_request` - A request to unsubscribe
+    ///
+    /// # Returns
+    ///
+    /// * [`UStatus`] detailing if unsubscription was successful and if not why not
+    async fn unsubscribe(&self, unsubscribe_request: UnsubscribeRequest) -> UStatus;
```
sophokles73, line 249:
```suggestion
    async fn unsubscribe(&self, unsubscribe_request: UnsubscribeRequest) -> Result<(), UStatus>;
```

src/core/usubscription.rs
```diff
@@ -0,0 +1,308 @@
...
+    ///
+    /// * `fetch_subscriptions_request` - A request to fetch subscriptions given a topic or subscriber
+    ///
+    /// # Returns
+    ///
+    /// * [`FetchSubscriptionsResponse`] detailing the zero or more subscriptions' info
+    async fn fetch_subscriptions(
+        &self,
+        fetch_subscriptions_request: FetchSubscriptionsRequest,
+    ) -> FetchSubscriptionsResponse;
```
sophokles73, line 263:
```suggestion
    ) -> Result<FetchSubscriptionsResponse, UStatus>;
```

src/core/usubscription.rs
```diff
@@ -0,0 +1,308 @@
...
+    ///
+    /// * `notifications_register_request` - A request to receive changes to subscription status
+    ///
+    /// # Returns
+    ///
+    /// * [`UStatus`] detailing if notification registration was successful and if not why not
+    async fn register_for_notifications(
+        &self,
+        notifications_register_request: NotificationsRequest,
+    ) -> UStatus;
```
sophokles73, line 278:
```suggestion
    ) -> Result<(), UStatus>;
```

src/core/usubscription.rs
```diff
@@ -0,0 +1,308 @@
...
+    ///
+    /// * `notifications_unregister_request` - A request to no longer receive changes to subscription status
+    ///
+    /// # Returns
+    ///
+    /// * [`UStatus`] detailing if notification unregistration was successful and if not why not
+    async fn unregister_for_notifications(
+        &self,
+        notifications_unregister_request: NotificationsRequest,
+    ) -> UStatus;
```
sophokles73, line 293:
```suggestion
    ) -> Result<(), UStatus>;
```

src/core/usubscription.rs
```diff
@@ -0,0 +1,308 @@
...
+    ///
+    /// * `fetch_subscribers_request` - Request containing topic for which we'd like all subscribers' info
+    ///
+    /// # Returns
+    ///
+    /// * [`FetchSubscriptionsResponse`] detailing subscriber info for the provided topic
+    async fn fetch_subscribers(
+        &self,
+        fetch_subscribers_request: FetchSubscribersRequest,
+    ) -> FetchSubscriptionsResponse;
```
sophokles73, line 307:
```suggestion
    ) -> Result<FetchSubscriptionsResponse, UStatus>;
```

src/core/usubscription.rs
PLeVasseur: Honestly, I went back and forth on whether to handle this the same way as we did with `UTransport`'s methods.

If you look into the methods which don't return a UStatus, but return something else, e.g. a `SubscriptionResponse` it contains already a UCode as well as more specific fields as to the current `SubscriptionStatus`'s state.

Perhaps I was not modelling this correctly in my mind here.

Hmm. If we separate out the conditions, let's say like this (abbreviated pseudo code, obvs):
```
if (SubscriptionRequest is malformed || SubscriptionRequest times out) {
    return Err(UStatus::fail_with_code(...));
} else {
    // where subscription_response is the payload back from the USubscription service
    return Ok(subscription_response);
}
```
then I can definitely see it being useful to be able to model those kinds of failure conditions which do not originate from the semantics of the USubscription service itself.

I'll go ahead and update! :+1:

PLeVasseur:
Hey @sophokles73 & @AnotherDaniel -- looking for your feedback on the [latest changes + comment](https://github.com/eclipse-uprotocol/up-rust/pull/83#discussion_r1563046125) :+1:

src/core/usubscription.rs
sophokles73: > Or, to put it another way, even if we fail to subscribe, if that failure is due to the USubscription service telling us that, we'd still return Ok(). Therefore, even if someone calls subscribe() and they get back a Ok(SubscriptionResponse) they must also peer into the SubscriptionResponse's SubscriptionStatus's UCode to ensure that it's UCode::OK.

And .... we are back at https://github.com/eclipse-uprotocol/up-spec/issues/112 :-)
But I still hope, that in a not too distant future, we will see the advent of service interfaces which contain the definition of the (successful) response message only and use a dedicated (different) error reporting mechanism ...

sophokles73 (approved):
LGTM

sophokles73:
@PLeVasseur do you want to squash this or should I do it when merging?

AnotherDaniel:
> Hey @sophokles73 & @AnotherDaniel -- looking for your feedback on the [latest changes + comment](https://github.com/eclipse-uprotocol/up-rust/pull/83#discussion_r1563046125) 👍

I'm good too - I wholeheartedly agree with the discussion about nested response-status, above - this is just a brain-breaker, especially coupled with the often unwieldy proto-generated types. But won't be solved in this PR, so...

src/core/usubscription.rs
PLeVasseur: I hear you... it's clunky atm imho

PLeVasseur:
> @PLeVasseur do you want to squash this or should I do it when merging?

Hey @sophokles73 -- went ahead and squashed it, good to merge :+1:

PR #90, opened 2024-04-18
UMessageBuilder::build() build UUID if unavailable

### LINKED ISSUE #88: UMesageBuilder::with_message_id()
The factory methods that build request(), response(), notificaiton(), and publish() should within the belly create the message ID as there is no real benefit to force the developer to call with_message_id(). 
Take a look at the up-java [UAttributeBuilder.java](https://github.com/eclipse-uprotocol/up-java/blob/main/src/main/java/org/eclipse/uprotocol/transport/builder/UAttributesBuilder.java), the factory methods when they instantiate the UAttributeBuilder object, the UUID is automatically created and populated. I don't think we need to add the flexibility to add different types of UUID, we're only supporting the one defined in the spec now.

PLeVasseur:
* Adds the ability for UMessageBuilder::build() to make use of UUIDBuilder::build() now that it's a singleton if the message_id was not provide.
* Updated doc comments around UMessageBuilder::build() to reflect the distinction.
* Highlighted in UMessageBuilder::with_message_id() that typically need not be used anymore, but there may be edge cases.

Fixes #88

### DIFF
--- a/src/umessage/umessagebuilder.rs
+++ b/src/umessage/umessagebuilder.rs
@@ -17,7 +17,7 @@ use protobuf::{Enum, EnumOrUnknown, Message};
 use crate::uattributes::{NotificationValidator, UAttributesError};
 use crate::{
     Data, PublishValidator, RequestValidator, ResponseValidator, UAttributes, UAttributesValidator,
-    UCode, UMessage, UMessageType, UPayload, UPayloadFormat, UPriority, UUri, UUID,
+    UCode, UMessage, UMessageType, UPayload, UPayloadFormat, UPriority, UUIDBuilder, UUri, UUID,
 };
 
 #[derive(Debug)]
@@ -107,13 +107,11 @@ impl UMessageBuilder {
     /// # Examples
     ///
     /// ```rust
-    /// use up_rust::{UMessageBuilder, UMessageType, UPayloadFormat, UPriority, UUIDBuilder, UUri};
-
+    /// use up_rust::{UMessageBuilder, UMessageType, UPayloadFormat, UPriority, UUri};
     ///
     /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
     /// let topic = UUri::try_from("my-vehicle/cabin/1/doors.driver_side#status")?;
     /// let message = UMessageBuilder::publish(topic.clone())
-    ///                    .with_message_id(UUIDBuilder::build())
     ///                    .build_with_payload("closed".into(), UPayloadFormat::UPAYLOAD_FORMAT_TEXT)?;
     /// assert_eq!(message.attributes.type_, UMessageType::UMESSAGE_TYPE_PUBLISH.into());
     /// assert_eq!(message.attributes.priority, UPriority::UPRIORITY_CS1.into());
@@ -142,14 +140,12 @@ impl UMessageBuilder {
     /// # Examples
     ///
     /// ```rust
-    /// use up_rust::{UMessageBuilder, UMessageType, UPayloadFormat, UPriority, UUIDBuilder, UUri};
-
+    /// use up_rust::{UMessageBuilder, UMessageType, UPayloadFormat, UPriority, UUri};
     ///
     /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
     /// let origin = UUri::try_from("my-vehicle/cabin/1/dashcam#event")?;
     /// let destination = UUri::try_from("my-cloud/companion/1/alarm")?;
     /// let message = UMessageBuilder::notification(origin.clone(), destination.clone())
-    ///                    .with_message_id(UUIDBuilder::build())
     ///                    .build_with_payload("unexpected movement".into(), UPayloadFormat::UPAYLOAD_FORMAT_TEXT)?;
     /// assert_eq!(message.attributes.type_, UMessageType::UMESSAGE_TYPE_NOTIFICATION.into());
     /// assert_eq!(message.attributes.priority, UPriority::UPRIORITY_CS1.into());
@@ -185,14 +181,12 @@ impl UMessageBuilder {
     /// # Examples
     ///
     /// ```rust
-    /// use up_rust::{UMessageBuilder, UMessageType, UPayloadFormat, UPriority, UUIDBuilder, UUri};
-
+    /// use up_rust::{UMessageBuilder, UMessageType, UPayloadFormat, UPriority, UUri};
     ///
     /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
     /// let method_to_invoke = UUri::try_from("my-vehicle/cabin/1/rpc.doors")?;
     /// let reply_to_address = UUri::try_from("my-cloud/dashboard/1/rpc.response")?;
     /// let message = UMessageBuilder::request(method_to_invoke.clone(), reply_to_address.clone(), 5000)
-    ///                    .with_message_id(UUIDBuilder::build())
     ///                    .build_with_payload("lock".into(), UPayloadFormat::UPAYLOAD_FORMAT_TEXT)?;
     /// assert_eq!(message.attributes.type_, UMessageType::UMESSAGE_TYPE_REQUEST.into());
     /// assert_eq!(message.attributes.priority, UPriority::UPRIORITY_CS4.into());
@@ -232,14 +226,14 @@ impl UMessageBuilder {
     ///
     /// ```rust
     /// use up_rust::{UMessageBuilder, UMessageType, UPayloadFormat, UPriority, UUIDBuilder, UUri};
-
     ///
     /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
     /// let invoked_method = UUri::try_from("my-vehicle/cabin/1/rpc.doors")?;
     /// let reply_to_address = UUri::try_from("my-cloud/dashboard/1/rpc.response")?;
     /// let request_id = UUIDBuilder::build();
+    /// // a service implementation would normally use
+    /// // `UMessageBuilder::response_for_request(&request_message.attributes)` instead
     /// let message = UMessageBuilder::response(reply_to_address.clone(), request_id.clone(), invoked_method.clone())
-    ///                    .with_message_id(UUIDBuilder::build())
     ///                    .build()?;
     /// assert_eq!(message.attributes.type_, UMessageType::UMESSAGE_TYPE_RESPONSE.into());
     /// assert_eq!(message.attributes.priority, UPriority::UPRIORITY_CS4.into());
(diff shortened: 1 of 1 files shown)


src/umessage/umessagebuilder.rs
```diff
@@ -108,12 +108,10 @@ impl UMessageBuilder {
     ///
     /// ```rust
     /// use up_rust::{UMessageBuilder, UMessageType, UPayloadFormat, UPriority, UUIDBuilder, UUri};
```
sophokles73, line 110: should probably remove the UUIDBuilder ...

src/umessage/umessagebuilder.rs
```diff
@@ -143,13 +141,11 @@ impl UMessageBuilder {
...
```
sophokles73, line 143: should probably remove the UUIDBuilder ...

src/umessage/umessagebuilder.rs
```diff
@@ -186,13 +182,11 @@ impl UMessageBuilder {
...
```
sophokles73, line 184: should probably remove the UUIDBuilder ...

src/umessage/umessagebuilder.rs
```diff
@@ -281,18 +273,16 @@ impl UMessageBuilder {
...
-
     ///
     /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// # let request_message_id = UUIDBuilder::build();
```
sophokles73, line 278: the request_message_id is part of the (visible) assertions below so it should not be fenced out FMPOV

src/umessage/umessagebuilder.rs
```diff
@@ -319,12 +309,10 @@ impl UMessageBuilder {
 
     /// Sets the message's identifier.
     ///
-    /// Every message must have an identifier and any of the build functions will return an error if no
-    /// identifier has been set.
+    /// Every message must have an identifier. If this function is not used, the identifier will
+    /// be stamped on the message during [`UMessageBuilder::build()`].
```
sophokles73, line 313: can we use
> If this function is not used, an identifier will be generated and set on the message ...

src/umessage/umessagebuilder.rs
```diff
@@ -425,13 +410,12 @@ impl UMessageBuilder {
     ///
     /// ```rust
     /// use up_rust::{UMessageBuilder, UMessageType, UPayloadFormat, UPriority, UUIDBuilder, UUri};
-
     ///
     /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// # let request_msg_id = UUIDBuilder::build();
```
sophokles73, line 415: should not be fenced out FMPOV

src/umessage/umessagebuilder.rs
```diff
@@ -461,14 +445,12 @@ impl UMessageBuilder {
     ///
     /// ```rust
     /// use up_rust::{UMessageBuilder, UMessageType, UPayloadFormat, UPriority, UUIDBuilder, UUri};
```
sophokles73, line 447: remove UUIDBuilder?

src/umessage/umessagebuilder.rs
```diff
@@ -500,13 +482,11 @@ impl UMessageBuilder {
     ///
     /// ```rust
     /// use up_rust::{UCode, UMessageBuilder, UPayloadFormat, UPriority, UUIDBuilder, UUri};
```
sophokles73, line 484: remove UUIDBuilder?

sophokles73 (changes requested):

src/umessage/umessagebuilder.rs
PLeVasseur: I'm thinking to make the assertion not visible, since it gives the wrong impression of what a user _should_ be doing.

They should not be plugging a newly created UUID into that slot, but instead using one they extracted from the Request message.

Another way to go about is to fence off creating a Request message using UMessageBuilder, then show how to destructure that to get its reqid.

Kinda like that route, will try.

src/umessage/umessagebuilder.rs
PLeVasseur: Same comment as [here](https://github.com/eclipse-uprotocol/up-rust/pull/90#discussion_r1571197600).

src/umessage/umessagebuilder.rs
```diff
@@ -231,21 +225,21 @@ impl UMessageBuilder {
...
     ///
     /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
     /// let invoked_method = UUri::try_from("my-vehicle/cabin/1/rpc.doors")?;
     /// let reply_to_address = UUri::try_from("my-cloud/dashboard/1/rpc.response")?;
-    /// let request_id = UUIDBuilder::build();
-    /// let message = UMessageBuilder::response(reply_to_address.clone(), request_id.clone(), invoked_method.clone())
-    ///                    .with_message_id(UUIDBuilder::build())
+    /// # let request_msg = UMessageBuilder::request(invoked_method.clone(), reply_to_address.clone(), 1000).build();
+    /// // extracting reqid from received Request message for use in building Response message
+    /// let reqid = request_msg.as_ref().unwrap().clone().attributes.as_ref().unwrap().id.as_ref().unwrap().clone();
```
sophokles73, line 235: I'm sorry, but FMPOV this only confuses the reader because
* the code won't compile if copied from docs to local workspace (no request_msg) and
* the code for extracting the reqid looks terrible ;-)

Can't we simply create a reqid using the UUIDBuilder and then add some wording indicating that a service implementation would usually use `response_for_request` instead? FMPOV this would make things much clearer and more explicit ...

PLeVasseur: I uhhh, will grant you point 2. 😅

Okay, I think you're right, having a copy-pasteable example is good and a comment can explain what we would normally do.

src/umessage/umessagebuilder.rs
```diff
@@ -225,21 +225,21 @@ impl UMessageBuilder {
...
     /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
     /// let invoked_method = UUri::try_from("my-vehicle/cabin/1/rpc.doors")?;
     /// let reply_to_address = UUri::try_from("my-cloud/dashboard/1/rpc.response")?;
-    /// # let request_msg = UMessageBuilder::request(invoked_method.clone(), reply_to_address.clone(), 1000).build();
-    /// // extracting reqid from received Request message for use in building Response message
-    /// let reqid = request_msg.as_ref().unwrap().clone().attributes.as_ref().unwrap().id.as_ref().unwrap().clone();
-    /// let response_msg = UMessageBuilder::response(reply_to_address.clone(), reqid.clone(), invoked_method.clone())
+    /// let request_id = UUIDBuilder::build();
+    /// // a service implementation would normally use
+    /// // `UUIDBuilder::response_for_request(&request_message.attributes)` instead
```
sophokles73, line 235: me personally, I would use `UMessageBuilder::response_for_request` ;-)

src/umessage/umessagebuilder.rs
```diff
@@ -413,15 +412,15 @@ impl UMessageBuilder {
...
     /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
     /// let invoked_method = UUri::try_from("my-vehicle/cabin/1/rpc.doors")?;
     /// let reply_to_address = UUri::try_from("my-cloud/dashboard/1/rpc.response")?;
-    /// # let request_msg = UMessageBuilder::request(invoked_method.clone(), reply_to_address.clone(), 1000).build();
-    /// // extracting reqid from received Request message for use in building Response message
-    /// let reqid = request_msg.as_ref().unwrap().clone().attributes.as_ref().unwrap().id.as_ref().unwrap().clone();
-    /// let message = UMessageBuilder::response(reply_to_address, reqid, invoked_method)
+    /// let request_msg_id = UUIDBuilder::build();
+    /// // a service implementation would normally use
+    /// // `UUIDBuilder::response_for_request(&request_message.attributes)` instead
```
sophokles73, line 422: see above

src/umessage/umessagebuilder.rs
```diff
@@ -523,16 +522,16 @@ impl UMessageBuilder {
...
     /// let invoked_method = UUri::try_from("my-vehicle/cabin/1/rpc.doors")?;
     /// let reply_to_address = UUri::try_from("my-cloud/dashboard/1/rpc.response")?;
-    /// # let request_msg = UMessageBuilder::request(invoked_method.clone(), reply_to_address.clone(), 1000).build();
-    /// // extracting reqid from received Request message for use in building Response message
-    /// let reqid = request_msg.as_ref().unwrap().clone().attributes.as_ref().unwrap().id.as_ref().unwrap().clone();
     /// let status = UCode::OK.value();
-    /// let message = UMessageBuilder::response(reply_to_address, reqid, invoked_method)
+    /// let request_msg_id = UUIDBuilder::build();
+    /// // a service implementation would normally use
+    /// // `UUIDBuilder::response_for_request(&request_message.attributes)` instead
```
sophokles73, line 533: and once more

src/umessage/umessagebuilder.rs
```diff
@@ -563,14 +562,13 @@ impl UMessageBuilder {
...
+    /// use up_rust::{UAttributes, UAttributesValidators, UMessageBuilder, UMessageBuilderError, UMessageType, UPriority, UUIDBuilder, UUri};
     ///
     /// let invoked_method = UUri::try_from("my-vehicle/cabin/1/rpc.doors").unwrap();
     /// let reply_to_address = UUri::try_from("my-cloud/dashboard/1/rpc.response").unwrap();
-    /// # let request_msg = UMessageBuilder::request(invoked_method.clone(), reply_to_address.clone(), 1000).build();
-    /// // extracting reqid from received Request message for use in building Response message
-    /// let reqid = request_msg.as_ref().unwrap().clone().attributes.as_ref().unwrap().id.as_ref().unwrap().clone();
-    /// let result = UMessageBuilder::response(reply_to_address, reqid, invoked_method)
+    /// // a service implementation would normally use
+    /// // `UUIDBuilder::response_for_request(&request_message.attributes)` instead
```
sophokles73, line 570: see above

src/umessage/umessagebuilder.rs
```diff
@@ -587,11 +585,10 @@ impl UMessageBuilder {
     /// # let lsb = UUIDBuilder::build().lsb;
     /// let invoked_method = UUri::try_from("my-vehicle/cabin/1/rpc.doors")?;
     /// let reply_to_address = UUri::try_from("my-cloud/dashboard/1/rpc.response")?;
-    /// # let request_msg = UMessageBuilder::request(invoked_method.clone(), reply_to_address.clone(), 1000).build();
-    /// // extracting reqid from received Request message for use in building Response message
-    /// let reqid = request_msg.as_ref().unwrap().clone().attributes.as_ref().unwrap().id.as_ref().unwrap().clone();
     /// let message_id = UUIDBuilder::build();
-    /// let message = UMessageBuilder::response(reply_to_address, reqid, invoked_method)
+    /// // a service implementation would normally use
+    /// // `UUIDBuilder::response_for_request(&request_message.attributes)` instead
```
sophokles73, line 590: see above

src/umessage/umessagebuilder.rs
```diff
@@ -698,15 +695,16 @@ impl UMessageBuilder {
...
     /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
     /// let invoked_method = UUri::try_from("my-vehicle/cabin/1/rpc.doors")?;
     /// let reply_to_address = UUri::try_from("my-cloud/dashboard/1/rpc.response")?;
-    /// # let request_msg = UMessageBuilder::request(invoked_method.clone(), reply_to_address.clone(), 1000).build();
-    /// // extracting reqid from received Request message for use in building Response message
-    /// let reqid = request_msg.as_ref().unwrap().clone().attributes.as_ref().unwrap().id.as_ref().unwrap().clone();
-    /// let message = UMessageBuilder::response(reply_to_address, reqid, invoked_method)
+    /// let request_id = UUIDBuilder::build();
+    /// // a service implementation would normally use
+    /// // `UUIDBuilder::response_for_request(&request_message.attributes)` instead
```
sophokles73, line 705: see above

src/umessage/umessagebuilder.rs
```diff
@@ -698,15 +695,16 @@ impl UMessageBuilder {
...
+    /// let message = UMessageBuilder::response(reply_to_address, request_id, invoked_method)
+    ///                    .with_message_id(UUIDBuilder::build())
```
sophokles73, line 707: any particular reason you want to set the message_id explicitly here?

src/umessage/umessagebuilder.rs
PLeVasseur: lmaooooooooooo :rofl:

src/umessage/umessagebuilder.rs
PLeVasseur: My bad -- got caught up in me reverting changes for those tests which do use the request id.

PLeVasseur:
> Okay @sophokles73 -- updated based on your feedback.
> 
> If you're good with it I think you can squash and merge.

All right, 2nd attempt.

Already squashed to one commit. If looks good we can merge.

sophokles73 (approved):
LGTM

PR #92, opened 2024-04-29
Support matching on source and sink address

sophokles73:
The UTransport trait has been changed to support registration of
listeners matching on both a message's source and sink addresses.

This addresses https://github.com/eclipse-uprotocol/up-spec/issues/73

### DIFF
--- a/src/utransport.rs
+++ b/src/utransport.rs
@@ -86,7 +86,7 @@ use crate::{UMessage, UStatus, UUri};
 /// ```
 #[async_trait]
 pub trait UListener: Send + Sync {
-    /// Performs some action on receipt of a message
+    /// Performs some action on receipt of a message.
     ///
     /// # Parameters
     ///
@@ -104,7 +104,7 @@ pub trait UListener: Send + Sync {
     /// or spawn it onto a new task and await there to allow current context to immediately continue.
     async fn on_receive(&self, msg: UMessage);
 
-    /// Performs some action on receipt of an error
+    /// Performs some action on receipt of an error.
     ///
     /// # Parameters
     ///
@@ -154,30 +154,30 @@ pub trait UTransport: Send + Sync {
     /// # Arguments
     ///
     /// * `source_filter` - The [source](`crate::UAttributes::source`) address pattern that the message to receive needs to match.
-    /// * `sink_filter` - The [sink](`crate::UAttributes::sink`) address pattern that the message to receive needs to match.
+    /// * `sink_filter` - The [sink](`crate::UAttributes::sink`) address pattern that the message to receive needs to match,
+    ///                   or `None` to indicate that the message must not contain any sink address.
     ///
     /// # Errors
     ///
     /// Returns an error if no message could be received, e.g. because no message matches the given addresses.
     async fn receive(
         &self,
-        source: Option<&UUri>,
-        sink: Option<&UUri>,
+        source_filter: &UUri,
+        sink_filter: Option<&UUri>,
     ) -> Result<UMessage, UStatus>;
 
-    /// Registers a listener to be called for Notification and RPC Request/Response messages destined to
-    /// a given address.
+    /// Registers a listener to be called for messages.
+    ///
+    /// The listener will be invoked for each message that matches the given source and sink filter patterns
+    /// according to the rules defined by the [UUri specification](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc).
     ///
     /// # Arguments
     ///
-    /// * `source_filter` - The [source](`crate::UAttributes::source`) address pattern that messages need to match,
-    ///                     or `None` to match all addresses.
+    /// * `source_filter` - The [source](`crate::UAttributes::source`) address pattern that messages need to match.
     /// * `sink_filter` - The [sink](`crate::UAttributes::sink`) address pattern that messages need to match,
-    ///                   or `None` to match all addresses.
-    /// * `listener` - The listener to invoke. The listener will be invoked for each message that matches the given
-    ///              source and sink filter patterns according to the rules defined by the
-    ///              [UUri specification](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc).
-    ///              The listener can be unregistered again using [`UTransport::unregister_listener`].
+    ///                   or `None` to match messages that do not contain any sink address.
+    /// * `listener` - The listener to invoke.
+    ///                The listener can be unregistered again using [`UTransport::unregister_listener`].
     ///
     /// # Errors
     ///
@@ -190,7 +190,7 @@ pub trait UTransport: Send + Sync {
     /// ```
     /// use std::sync::Arc;
     /// use up_rust::UListener;
-    /// # use up_rust::{UMessage, UMessageType, UStatus, UTransport, UUri};
+    /// # use up_rust::{UMessage, UStatus, UTransport, UUri};
     /// # use async_trait::async_trait;
     /// #
     /// # pub struct MyTransport;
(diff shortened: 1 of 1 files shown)


sophokles73:
This should support all use cases ranging from implementing service consumers/providers to uStreamers.

WDYT? @PLeVasseur @evshary @AnotherDaniel

src/utransport.rs
```diff
@@ -331,51 +350,41 @@ mod tests {
...
-    use std::collections::hash_map::Entry;
     use std::collections::{HashMap, HashSet};
-    use std::sync::{Arc, Mutex};
+    use std::sync::Arc;
 
+    #[derive(Default)]
     struct UPClientFoo {
-        #[allow(clippy::type_complexity)]
-        listeners: Arc<Mutex<HashMap<UUri, HashSet<ComparableListener>>>>,
+        source_listeners: HashMap<UUri, HashSet<ComparableListener>>,
```
PLeVasseur, line 358: :+1:

src/utransport.rs
```diff
@@ -230,38 +240,47 @@ pub trait UTransport: Send + Sync {
...
-    /// let unregister_result = my_transport.unregister_listener(my_uuri, my_listener);
+    /// my_transport.unregister_listener(
+    ///     Some(&source_uuri),
+    ///     None,
+    ///     my_listener);
     /// ```
     async fn register_listener(
-        &self,
-        topic: UUri,
+        &mut self,
```
PLeVasseur, line 260: Why is an exclusive reference needed instead of a shared reference in the new formulation?

src/utransport.rs
```diff
@@ -230,38 +240,47 @@ pub trait UTransport: Send + Sync {
...
+    /// * `sink` - The sink address pattern that the listener had been registered for.
+    /// * `listener` - The listener to unregister.
     ///
     /// # Errors
     ///
     /// Returns an error if the listener could not be unregistered, for example if the given listener does not exist.
     async fn unregister_listener(
-        &self,
-        topic: UUri,
+        &mut self,
```
PLeVasseur, line 281: Same question as up [here](https://github.com/eclipse-uprotocol/up-rust/pull/92/files#r1583558975), but:

Why is an exclusive reference needed instead of a shared reference in the new formulation?

src/utransport.rs
```diff
@@ -180,7 +190,7 @@ pub trait UTransport: Send + Sync {
     /// ```
     /// use std::sync::Arc;
     /// use up_rust::UListener;
-    /// # use up_rust::{UMessage, UStatus, UTransport, UUri};
+    /// # use up_rust::{UMessage, UMessageType, UStatus, UTransport, UUri};
```
PLeVasseur, line 193: Is `UMessageType` used?

src/utransport.rs
```diff
@@ -230,38 +240,47 @@ pub trait UTransport: Send + Sync {
...
+    ///     Some(&source_uuri),
+    ///     None,
+    ///     my_listener);
     /// ```
     async fn register_listener(
-        &self,
-        topic: UUri,
+        &mut self,
+        source_filter: Option<&UUri>,
+        sink_filter: Option<&UUri>,
```
PLeVasseur, line 262: If I'm understanding correctly, the idea is rather than discriminate based on an explicit message type, to instead allow for implicit selection among the different types based on source and sink address.

I wanted to come back to something we discussed -- discriminating any Publish from any Notification.

I understand that the Notification will have a sink UUri.

Could you flesh out how we'd describe
1. "give me all Notifications (i.e. all messages whose sink is not empty)" vs 
2. "give me all Publishes (i.e. all messages whose sink is empty)"? 

Perhaps 1. we'd provide an empty UUri indicating any UUri filled will match? I'm not quite sure how we'd filter to get 2. as the semantics outlined is None => matches all UUris (including no UUri?).

src/utransport.rs
```diff
@@ -153,21 +153,31 @@ pub trait UTransport: Send + Sync {
...
-    /// Registers a listener to be called for each message that is received on a given address.
+    /// Registers a listener to be called for Notification and RPC Request/Response messages destined to
+    /// a given address.
     ///
     /// # Arguments
     ///
-    /// * `address` - The (resolved) address to register the listener for.
-    /// * `listener` - The listener to invoke. Note that we do not take ownership to communicate
-    ///                that a caller should keep a copy to be able to call `unregister_listener()`
+    /// * `source_filter` - The [source](`crate::UAttributes::source`) address pattern that messages need to match,
```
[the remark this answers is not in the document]
devkelley: I would agree with your thoughts that some kind of partial match is needed to allow for wildcarding. I would restrict partial matching to the authority level only though.

src/utransport.rs
devkelley: I think using an empty UUri to indicate any address would work. and then `None` truly means no UURI. I agree that there needs to be a distinction between `any` uuri and `no` uuri

src/utransport.rs
```diff
@@ -230,38 +240,47 @@ pub trait UTransport: Send + Sync {
...
+    /// my_transport.unregister_listener(
+    ///     Some(&source_uuri),
+    ///     None,
+    ///     my_listener);
     /// ```
     async fn register_listener(
-        &self,
-        topic: UUri,
+        &mut self,
+        source_filter: Option<&UUri>,
```
evshary, line 261: Is there any case we have None in source UUri? I guess we always need source UUri.

src/utransport.rs
evshary: > I think using an empty UUri to indicate any address would work. and then None truly means no UURI. I agree that there needs to be a distinction between any uuri and no uuri

I second it. Perhaps we should also put the definition that empty UUri indicates any address into spec.

src/utransport.rs
evshary: I think this is because there is no Mutex here, so needs to be mutable to modify the hash table.
https://github.com/eclipse-uprotocol/up-rust/pull/92/files#diff-2b11318fb2673b25403b992883ee6ac91316e6f11f3c32c1b352caf4e07c79beR358

sophokles73: yes, that is the reason. Since un-/registering will (generally speaking) require some modification of the transport's state, I do not think that this is an issue but actually provides more options for implementation (as we see in the unit tests). WDYT?

src/utransport.rs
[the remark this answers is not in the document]
sophokles73: That is exactly what I had in mind. This is also the definition of UUri matching in my recent [UUri Spec PR](https://github.com/eclipse-uprotocol/up-spec/pull/121)

src/utransport.rs
sophokles73: I guess not

src/utransport.rs
sophokles73: @PLeVasseur I see your point.

This is very close to what I already specified in my UUri PR, i.e. using empty property values in a UUri to match any value.

@devkelley I believe this could work.

@evshary I will amend my UUri PR accordingly.

src/utransport.rs
sophokles73: That is probably true and if we agree to use _empty property values_ in a UUri to match any value, then I guess we could change this to just `&UUri`.

src/utransport.rs
PLeVasseur: > I am not sure if you really meant _Notification_ type messages in your first example or if you actually meant _all non-Publish messages_?

I meant all non-Publish, must have gotten mixed up.

Your examples makes sense. 👍

devkelley (approved):

src/utransport.rs
evshary: I'm fine with that. We can use `Arc` + `Mutex` to `UpClientXXX` if need to pass to other threads.

evshary (approved):
LGTM. Perhaps it would be better to merge the PR after discussing it with other commiters.

src/utransport.rs
sophokles73: Ok, it turns out that using `&mut self` here is not such a great idea because it requires client code to get a grip on a mutable `Arc<dyn UTransport>` which is not as easy as it sounds. The [Arc::get_mut documentation](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.get_mut) tells us why. I will change this back to just `&self` in order to make implementation of client code easier, even if this means some additional effort for implementors of UTransport.

PR #98, opened 2024-05-08
Fix USubscription doc tests

sophokles73:
The doc tests had not been updated yet to reflect the changes
introduced to the UTransport trait's functions for matching on
source and sink address patterns.

### DIFF
--- a/src/core/usubscription.rs
+++ b/src/core/usubscription.rs
@@ -49,15 +49,15 @@ use async_trait::async_trait;
 /// #             todo!()
 /// #         }
 /// #
-/// #         async fn receive(&self, topic: UUri) -> Result<UMessage, UStatus> {
+/// #         async fn receive(&self, source_filter: &UUri, sink_filter: Option<&UUri>) -> Result<UMessage, UStatus> {
 /// #             todo!()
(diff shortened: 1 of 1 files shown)


PLeVasseur (approved):
LGTM

@sophokles73 -- does that mean the doc tests didn't catch this and fail the build when the `UTransport::register_listener()` change went in?

sophokles73:
> LGTM
> 
> @sophokles73 -- does that mean the doc tests didn't catch this and fail the build when the `UTransport::register_listener()` change went in?

At least not as part of the tests we run in our standard CI build. That is because that standard build does not activate the `usubscription` feature, but the _nightly_ build does :-)

AnotherDaniel:
This is interesting insofar as I'd expect the --all-features flag to enable, well, all features?

sophokles73:
> This is interesting insofar as I'd expect the --all-features flag to enable, well, all features?

@AnotherDaniel 
Doc tests are not (yet) supported in `cargo nextest`:
https://github.com/nextest-rs/nextest/issues/16

Maybe you can add an extra `cargo test --doc` step to the corresponding job?

Issue #84, opened 2024-04-11
Build pipeline evolution


AnotherDaniel:
... to collect next steps in the evolution of our build pipeline and associated workflows... (and also so we can close #51 )

- [ ] review the scripts in ./tools, ensure we have locally-runnable versions especially of the check workflow
- [ ] #73 add x-compilation from x86 to arm, to go into our nightly workflow and mainly serve as a proof-point/canary which I'm not sure we'll ever see roll off the perch
- [X] #25 incorporate the EF Dash tool for code scanning
- [ ] add cross-compile support for [all valid Android targets](https://doc.rust-lang.org/rustc/platform-support/android.html)
- [X] add doctest steps to check and nightly workflows (as nexttest doesn't do doctests atm)

PR #101, opened 2024-05-14
Add doctests (as nextest doesn't do these yet)

AnotherDaniel:
As has [come up recently](https://github.com/eclipse-uprotocol/up-rust/pull/98#issuecomment-2106745523), our check workflow doesn't yet run doctests. So this is adding an explicit doctest step.

(refer to last bullet in #84)

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -31,6 +31,9 @@ on:
       test_results_url:
         description: "URL of the test results artifact"
         value: ${{ jobs.nextest.outputs.test_results_url }}
+      doctest_results_url:
+        description: "URL of the doctest results artifact"
+        value: ${{ jobs.doctest.outputs.test_results_url }}
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yaml
```diff
@@ -147,4 +147,25 @@ jobs:
         with:
           name: test-results
           path: testresults--all-features.json
-          
\ No newline at end of file
+
+  doctest:
+    # Run doctests separately, as nextest doesn't yet (https://github.com/nextest-rs/nextest/issues/16)
+    outputs:
+      test_results_url: ${{ steps.doctest_results.outputs.artifact-url }}
```
sophokles73, line 154: any reason why this is not added to the workflow's outputs (as the output from nextest)?

.github/workflows/check.yaml
```diff
@@ -147,4 +147,25 @@ jobs:
...
+    env:
+      RUSTDOCFLAGS: -Dwarnings
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@master
+        with: 
+          toolchain: ${{ env.RUST_TOOLCHAIN }}
+      - name: Run doc tests
+        run: |
+          cargo test --doc --all-features
```
sophokles73, line 165: if we create JSON output from nextest, maybe we also want to create JSON output here?

.github/workflows/check.yaml
AnotherDaniel: Yeah - it's already been to long since I did all this; fixed.

.github/workflows/check.yaml
```diff
@@ -147,4 +150,26 @@ jobs:
...
+    env:
+      RUSTDOCFLAGS: -Dwarnings
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@master
+        with: 
+          toolchain: ${{ env.RUST_TOOLCHAIN }}
+      - name: Run doc tests
+        run: |
+          cargo test --doc --all-features > doctestresults--all-features.json
```
sophokles73, line 168: this will not produce JSON output
for that, you need to provide some command line params:
```sh
RUSTC_BOOTSTRAP=1 cargo test --doc -- -Z unstable-options --format json --report-time > doctestresults--all-features.json
```

.github/workflows/check.yaml
```diff
@@ -147,4 +150,26 @@ jobs:
...
+      - name: Upload doctest results artifact
+        id: doctest_results
+        uses: actions/upload-artifact@v4
+        with:
+          name: doctest-results
+          path: doctestresults--all-features.json
+
```
sophokles73, line 175: missing EOL

.github/workflows/check.yaml
AnotherDaniel: I really shouldn't be doing this while being at a conference...

sophokles73 (approved):
LGTM

PR #102, opened 2024-05-14
Remove support for CloudEvents

### LINKED ISSUE #100: CloudEvent module seems to be useless
It is unclear what the CloudEvent module's functionality would be useful for.
All uProtocol APIs revolve around UMessages and it is up to a transport implementation if it wants to map UMessages to CloudEvents when writing the messages to the transport PDUs.

The code is also pretty much unmaintainable with the hundreds of verbatim unit tests ...

I propose to simply remove the module and add something more suitable once we see the need for it.

@AnotherDaniel @stevenhartley WDYT?

sophokles73:
The functionality implemented by the cloudevents module did not serve
any real purpose since the introduction of UMessage as the primary
means to pass messages around. UTransport implementations may
or may not choose to map UMessages to CloudEvents before writing to
the transport PDU. The mapping is part of the uProtocol Specification
but the cloudvents module did not specifically support it.

Therefore, CloudEvent support has been removed from up-rust.

Fixes #100


AnotherDaniel (approved):
Getting rid of a bunch of dependencies - I like it!

PR #104, opened 2024-05-15
Adopt recent changes to Core API

sophokles73:
The recent changes made to uri.proto, upayload.proto and uattributes.proto have been incorporated into the up-rust library.

Along with these changes, the library now uses the proto3 files from the `main` branch of the _up-spec_ repository.


AnotherDaniel (approved):
Massive changes for sure. But as it's all aligned with the core URI rework, LGTM

PR #105, opened 2024-05-15
Verify encoding of Notification message type

sophokles73:
When mapping a UMessage to a transport PDU, the message type needs
to be encoded to a string. The UMessageType struct's unit tests have
been extended to verify that this also works for the Notification
message type.


AnotherDaniel (approved):
LGTM

PR #108, opened 2024-05-17
Remove UPayload

AnotherDaniel:
Following [up-spec #145](https://github.com/eclipse-uprotocol/up-spec/pull/145), remove UPayload from the rust library. This let's us get rid of RpcMapper entirely, which is nice. For now I did not 'save' the `RpcMapper::pack_any()` and `RpcMapper::unpack_any()` methods, because they are not needed anywhere I can see and also have little to do with the actual uProtocol scope. Shout if you're missing them nonetheless!


src/rpc/rpcclient.rs
```diff
@@ -47,7 +47,7 @@ pub trait RpcClient: Send + Sync {
     async fn invoke_method(
         &self,
         method: UUri,
-        request: UPayload,
+        request: UMessage,
```
sophokles73, line 50: if we accept a UMessage as parameter, then there is no need for priority, ttl and token anymore because these will be contained in UMessage::attributes.

src/umessage/umessagebuilder.rs
```diff
@@ -713,19 +669,15 @@ impl UMessageBuilder {
...
-                self.build_with_payload(
-                    serialized_payload.into(),
-                    UPayloadFormat::UPAYLOAD_FORMAT_PROTOBUF,
-                )
-            })
+    ) -> Result<UMessage, UMessageError> {
+        self.payload_format = UPayloadFormat::UPAYLOAD_FORMAT_PROTOBUF;
+        let mut message = self.build()?;
+        message
+            .insert_protobuf_payload(payload, Some(self.payload_format))
```
sophokles73, line 679: it feels very strange that we use the builder to build the message and then _alter_ the message ...
Is there a particular reason why this builder's build functions cannot remain as they were? Why do we need `Message::insert_protobuf_payload`?

src/umessage.rs
```diff
@@ -17,3 +17,151 @@ mod umessagetype;
...
+    /// * `payload_format`: Optional `UMessage::UPayloadFormat`, defining serialization to use. Will default to UPayloadFormat::UPAYLOAD_FORMAT_PROTOBUF.
+    ///
+    /// # Returns
+    ///
+    /// * Ok(()): if the operation succeeded
+    ///
+    /// # Errors
+    ///
+    /// * Err(`UMessageError`) if the serialization process fails, for example if the payload format is unsupported.
+    pub fn insert_protobuf_payload<T: MessageFull>(
```
sophokles73, line 90: FMPOV we do not need this function and should instead keep the builder's corresponding functionality. Or am I missing something?

AnotherDaniel: The thought was that it would be useful to modify the payload of an existing UMessage object - so these methods went into UMessage, not least because they are mirroring the extract_payload() methods that imo definitely belong in UMessage.

If that is not a valid use case at all, then I agree and the logic can move into UMessageBuilder.

PLeVasseur: IMHO even though we're not forbidding anyone from modifying a `UMessage`, since we're publicly exposing the protobuf type, I like the idea of making the "obvious" way to make a `UMessage` very much a one-way-of-doing-things process.

So my stance would be that we should not go out of our way to support inserting a protobuf payload after the fact, folks should make use of the `UMessageBuilder` to just make the message they need with the payload they need.

> not least because they are mirroring the extract_payload() methods that imo definitely belong in UMessage.

Sorry, I'm not familiar with what you're writing about here. I agree there should be some "nice" way of getting at the payload.

AnotherDaniel: Ok, bowing to majority opinion - moved the code to UMessageBuilder.

evshary (approved):
LGTM.
Just one question: Should we pin at a specific commit of up-spec instead of main in `build.rs`? When the up-spec (up-core-api) changes, it'll break the up-rust and affect other libraries depending on up-rust.
I think it'll be better to fix it at a specific commit and keep updating regularly.

sophokles73:
> Just one question: Should we pin at a specific commit of up-spec instead of main in `build.rs`? When the up-spec (up-core-api) changes, it'll break the up-rust and affect other libraries depending on up-rust.

Actually, that is the main intention of referring to `up-spec/main`, i.e. we _want_ to see that things have changed and we need to adapt up-rust ...

src/umessage/umessagebuilder.rs
```diff
@@ -70,7 +37,7 @@ pub struct UMessageBuilder {
     permission_level: Option<u32>,
     comm_status: Option<EnumOrUnknown<UCode>>,
     request_id: Option<UUID>,
-    payload: Option<Bytes>,
+    payload: Option<Vec<u8>>,
```
sophokles73, line 40: what is the reason for changing this? FMPOV Bytes is much more flexible when it comes to writing/reading and does require much less copying ...

src/umessage/umessagebuilder.rs
```diff
@@ -713,19 +673,39 @@ impl UMessageBuilder {
...
-            .write_to_bytes()
-            .map_err(UMessageBuilderError::from)
-            .and_then(|serialized_payload| {
-                self.build_with_payload(
-                    serialized_payload.into(),
-                    UPayloadFormat::UPAYLOAD_FORMAT_PROTOBUF,
-                )
-            })
+    ) -> Result<UMessage, UMessageError> {
+        if !(self.payload_format == UPayloadFormat::UPAYLOAD_FORMAT_PROTOBUF
```
sophokles73, line 680: where is `self.payload_format` set? Before your change, the payload format always was `UPayloadFormat::UPAYLOAD_FORMAT_PROTOBUF`. If we now want to also support `UPayloadFormat::UPAYLOAD_FORMAT_PROTOBUF_WRAPPED_IN_ANY`, then we either need to pass in the payload format as a param, add a dedicated function for building with wrapped payload or detect if the passed in payload actually is an `Any`. FMPOV we should go with option 2, making the whole thing very explicit and also making it very easy to later deprecate and remove support for it again ...

sophokles73 (changes requested):

sophokles73:
@evshary Would you mind taking a look at the changes that I have pushed in addition to Daniel's original PR? @AnotherDaniel is on vacation this week and we agreed to get this merged this week ...

sophokles73 (approved):
LGTM

sophokles73:
@evshary can you check if this (still) works for you?

src/umessage/umessagebuilder.rs
```diff
@@ -362,7 +368,14 @@ impl UMessageBuilder {
         {
             assert!(priority.value() >= UPriority::UPRIORITY_CS4.value())
         }
-        self.priority = priority;
+        if priority != PRIORITY_DEFAULT {
```
evshary, line 371: Does that mean if the priority is default, then we don't need to include the priority in the serialized protobuf?

evshary: What if priority equals to `UPriority::UPRIORITY_UNSPECIFIED`?

sophokles73: > Does that mean if the priority is default, then we don't need to include the priority in the serialized protobuf?

According to https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/qos.adoc I would say so, yes.

evshary: I kind of got your point.

evshary:
> @evshary can you check if this (still) works for you?

Yes, I give my pass to it.

src/umessage/umessagebuilder.rs
sophokles73: > What if priority equals to `UPriority::UPRIORITY_UNSPECIFIED`?

Then the priority will not be included in the serialized protobuf because it has the default value (0 in this case).

PR #111, opened 2024-05-23
Fix build.rs script

sophokles73:
The proto3 files from up-spec have been change to include version
information which has caused the build.rs script to fail to find the
included files. This has been fixed.


evshary (approved):
LGTM

PR #112, opened 2024-05-24
Use custom UMessage struct

sophokles73:
The build.rs script no longer pulls in the umessage.proto from up-spec.
Instead, the umessage module no defines the UMessage struct with
private fields so that UMessages can now only be created (correctly) via
the UMessageBuilder.

### DIFF
--- a/src/umessage.rs
+++ b/src/umessage.rs
@@ -14,11 +14,10 @@
 mod umessagebuilder;
 mod umessagetype;
 
+use bytes::Bytes;
 pub use umessagebuilder::*;
 
-pub use crate::up_core_api::umessage::UMessage;
-
-use crate::{UAttributesError, UPayloadFormat};
+use crate::{UAttributes, UAttributesError, UPayloadFormat};
 use protobuf::{well_known_types::any::Any, Message};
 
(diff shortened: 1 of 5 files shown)


src/umessage.rs
```diff
@@ -63,7 +62,22 @@ impl From<&str> for UMessageError {
...
 
+/// A container for a message's attributes and payload.
+#[derive(Debug, Clone)]
+pub struct UMessage {
+    attributes: UAttributes,
+    payload: Option<Bytes>,
+}
+
 impl UMessage {
+    pub fn attributes(&self) -> &UAttributes {
```
PLeVasseur, line 73: I really like this!

So the idea is that we have now not generated the code for the UMessage .proto object, so we can _force_ correctness, as the only way to build a UMessage is via UMessageBuidler and we've made the struct fields private and can only grab the UAttributes and payload Bytes as reference.

Slick!!

sophokles73: That _would have been_ the idea, yes. Sadly, though, UMessage is actually being used by uTwin in its input/output messages, so the initial approach will not work ...

PR #113, opened 2024-05-24
Fix validation of Notification URI

sophokles73:
The destination (sink) address of a Notification message must
contain a resource ID of 0.

The corresponding checks have been updated/added.


evshary (approved):
LGTM

PR #114, opened 2024-05-27
Fix the path of building protobuf

evshary:
The import path is changed after https://github.com/eclipse-uprotocol/up-spec/pull/161, so we need to keep `uprotocol` while building proto.


evshary:
Close after https://github.com/eclipse-uprotocol/up-rust/pull/115

PR #116, opened 2024-05-28
Implement missing requirements from UUri specification

sophokles73:
* Added pattern matching logic
* Added missing checks for parsing URI strings
* Added OpenFastTrace tags to source code

### DIFF
--- a/src/uri.rs
+++ b/src/uri.rs
@@ -11,8 +11,13 @@
  * SPDX-License-Identifier: Apache-2.0
  ********************************************************************************/
 
+// [impl->dsn~data-model-naming~1]
+// [impl->req~data-model-proto~1]
+
 use std::hash::{Hash, Hasher};
 use std::str::FromStr;
+
+use uriparse::URIReference;
 
 pub use crate::up_core_api::uri::UUri;
 
@@ -57,6 +62,7 @@ impl std::fmt::Display for UUriError {
 
 impl std::error::Error for UUriError {}
 
+// [impl->req~uri-serialization~1]
 impl From<&UUri> for String {
     /// Serializes a UUri to a URI string.
     ///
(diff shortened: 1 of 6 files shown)


src/uri.rs
```diff
@@ -102,86 +112,90 @@ impl FromStr for UUri {
...
+    // [impl->dsn~uri-scheme~1]
+    // [impl->dsn~uri-host-only~1]
+    // [impl->dsn~uri-authority-mapping~1]
+    // [impl->dsn~uri-path-mapping~1]
+    // [impl->req~uri-serialization~1]
     fn from_str(uri: &str) -> Result<Self, Self::Err> {
         if uri.is_empty() {
             return Err(UUriError::serialization_error("URI is empty"));
         }
+        let parsed_uri = URIReference::try_from(uri)
```
evshary, line 125: Could uriparse deal with UURI w/o authority?
For example: https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc#31-examples

src/uri.rs
```diff
@@ -102,86 +112,90 @@ impl FromStr for UUri {
...
+            ));
+        }
+        if parsed_uri.has_fragment() {
             return Err(UUriError::serialization_error(
-                "Remote URI must contain authority, entity ID, entity version and resource ID",
+                "uProtocol URI must not contain fragment",
             ));
-        } else if uri_parts[2].trim().is_empty() {
+        }
+        let authority_name = parsed_uri
```
evshary, line 145: It suddenly comes to my mind: Is it possible to have the asterisk in the authority name?
I mean `up//*/...` means matching all authority name, but how about `up/auth*/...`?

src/uri.rs
sophokles73: yes, that is one of the test cases https://github.com/eclipse-uprotocol/up-rust/blob/01e3149aa7ebffeb81f03746791693f198076c5e/src/uri.rs#L549

src/uri.rs
sophokles73: It is (currently) possible to use an authority that includes an asterisk, but that would have no special meaning, i.e. it would not be recognized nor used as a wildcard when matching.

evshary (approved):
LGTM

AnotherDaniel (approved):
Like the change to an existing uri parsing crate, LGTM

evshary:
Allow me to merge the PR. I need this in my up-client-zenoh-rust implementation :smiley:

Issue #122, opened 2024-06-17
Implement Communication Level API (L2)


sophokles73:
The [uP-L2 API](https://github.com/eclipse-uprotocol/up-spec/blob/main/up-l2/api.adoc) needs to be implemented based on the uP-L1 API

sophokles73:
closed via #147

PR #124, opened 2024-06-18
Add abstract Communication Layer API traits

sophokles73:
Added traits for publishing messages, sending notifications and
invoking remote service operations. These traits represent the
uProtocol Communication Layer API that uEntities can use to
interact with each other using uProtocol.

This is the first step for #122


src/communication/rpc.rs
```diff
@@ -0,0 +1,87 @@
...
+    async fn invoke_method(&self, request: UMessage) -> Result<UMessage, ServiceInvocationError>;
+}
+
+/// A server for exposing RPC endpoints.
+///
+/// Please refer to the
+/// [Communication Layer API Specifications](https://github.com/eclipse-uprotocol/up-spec/blob/main/up-l2/api.adoc).
+#[async_trait]
+pub trait RpcServer: Send + Sync {
+    async fn register_endpoint(
```
AnotherDaniel, line 75: Just in keeping with the other method declarations so far: do we want some docstrings here?

src/communication/rpc.rs
```diff
@@ -0,0 +1,87 @@
...
+        &self,
+        source_filter: &UUri,
+        method: &UUri,
+        listener: Arc<dyn RpcClient>,
+    ) -> Result<(), RegistrationError>;
+    async fn unregister_endpoint(
```
AnotherDaniel, line 81: Same comment as for register_endpoint()

src/utransport.rs
```diff
@@ -16,9 +16,23 @@ use std::hash::{Hash, Hasher};
...
 
-use crate::{UMessage, UStatus, UUri};
+use crate::{UCode, UMessage, UStatus, UUri};
 
-/// `UListener` is the uP-L1 interface that provides a means to create listeners which are registered to [`UTransport`]
+/// A factory for URIs representing this uEntity's resources.
+///
+/// Implementations may use arbitrary mechanisms to determine the information that
+/// is necessary for creating URIs, e.g. environment variables, configuration files etc.
+pub trait LocalUriProvider: Send + Sync {
```
AnotherDaniel, line 25: Is there a specific usage scenario that we're aware of, where this trait needs to be Send+Sync?

AnotherDaniel (commented):
One suggestion, one question...

src/utransport.rs
sophokles73: I assume that an implementation of this trait will be passed into an implementation of e.g. RpcClient. If we want RpcClient to be used from different threads concurrently, then LocalUriProvider being Send + Sync will allow the RpcClient to simply invoke its functions without bothering about synchronization.

AnotherDaniel: Ok thanks. I will implement this asap, because I've got equivalent code in my usub project already and want to see what this looks like from that perspective.

AnotherDaniel (approved):
LGTM

src/utransport.rs
```diff
@@ -250,16 +273,23 @@ pub trait UTransport: Send + Sync {
     /// ```
     async fn register_listener(
```
evshary, line 274: I'm thinking the reason you only have the default `UNIMPLEMENTED` error on `receive` and `register_listener` but not the `send`. Is this because developers can choose to implement either `receive` or `register_listener`?

src/utransport.rs
```diff
@@ -16,9 +16,23 @@ use std::hash::{Hash, Hasher};
...
-/// `UListener` is the uP-L1 interface that provides a means to create listeners which are registered to [`UTransport`]
+/// A factory for URIs representing this uEntity's resources.
+///
+/// Implementations may use arbitrary mechanisms to determine the information that
+/// is necessary for creating URIs, e.g. environment variables, configuration files etc.
+pub trait LocalUriProvider: Send + Sync {
+    /// Gets the _authority_ used for URIs representing this uEntity's resources.
+    fn get_authority(&self) -> String;
+    /// Gets a URI that represents a given resource of this uEntity.
+    fn get_resource(&self, resource_id: u16) -> UUri;
```
evshary, line 29: Naming `get_resource` misled me this will return `resource_id`. Since I guess the function is only used in Publish, how about `get_publish_uuri` or something similar?

src/utransport.rs
evshary, line 25: Double-check whether I have the right understanding. The UPTransport should have its own ways (no matter env, config, constructor...) to initialize the `authority`, `entity_id`, and `version`. Then we can take the benefit of `LocalUriProvider` after implementing the trait, right?

I only have `authority` in `up-transport-zenoh-rust` now. I will also support getting `entity_id` and `version` from users.

evshary (approved):
Add some questions
LGTM basically.

src/utransport.rs
sophokles73: exactly, or both, actually

src/utransport.rs
sophokles73: It will also be needed for registering an RPC server endpoint. What about `get_resource_uri`?

src/utransport.rs
sophokles73: You got it exactly right. Does this (still) make sense to you?

src/utransport.rs
evshary: Sound good to me.

src/utransport.rs
evshary: Yes, make sense to me.

src/communication/rpc.rs
sophokles73: Yes, definitely. Another thing I stumbled upon: for listeners to Publish messages it makes sense to be able to register multiple listeners for the same source address. However, I wonder if we should restrict the number of request listeners per (local) resource ID to one only. Otherwise you might end up in a situation where a client gets back multiple responses to its request. WDYT? @AnotherDaniel @evshary

AnotherDaniel: Yes, agree - also this would give implementors on server-side a little less rope to hang themselves with...

evshary: Second it. I remember we had a conclusion that Notification and RPC should be 1-to-1, and Publish is 1-to-N in the weekly meeting. I think it will be more simplified.

PR #125, opened 2024-06-19
Migrate UUIDBuilder to use UUIDv7

sophokles73:
The UUIDBuilder now creates v7 UUIDs as defined by RFC 4122 using random bits only, i.e. no sub-millisecond precision info nor other monotonic counters.


evshary (approved):
LGTM

sophokles73:
> I wonder if, at this point, we even need a separate UUIDbuilder any more? Might it make sense to just move that one left-over construction method to uuid.rs?

I had the same thought, but that would break the API ... WDYT?

AnotherDaniel:
> > I wonder if, at this point, we even need a separate UUIDbuilder any more? Might it make sense to just move that one left-over construction method to uuid.rs?
> 
> I had the same thought, but that would break the API ... WDYT?

I think that at the moment (the L2 changes) we've already caused sufficient API havoc that this little break won't really make a difference ;-)

sophokles73:
> > > I wonder if, at this point, we even need a separate UUIDbuilder any more? Might it make sense to just move that one left-over construction method to uuid.rs?
> > 
> > 
> > I had the same thought, but that would break the API ... WDYT?
> 
> I think that at the moment (the L2 changes) we've already caused sufficient API havoc that this little break won't really make a difference ;-)

@evshary @PLeVasseur would you also be fine with moving the functionality for creating a UUID into the `UUID` type and remove the `UUIDBuilder` altogether?

PLeVasseur (approved):
Makes sense to remove UUIDBuilder with the change to use UUIDv7. Less bespoke domain logic.

PR #126, opened 2024-06-20
Use custom parameter types in L2 API

sophokles73:
In order to better decouple uEntities from the L1 (proto defined)
types, some use-case specific parameter types have been added.

This is for #122


evshary (approved):
LGTM

PR #127, opened 2024-06-21
Add missing usubscription type export, minor qol functions

AnotherDaniel:
Minor items:
- Missing export of UnsubscribeResponse type in usubscription module
- is_empty() checks for UUri and SusbcriberInfo types (will compare with ::default() objects)

### DIFF
--- a/src/communication.rs
+++ b/src/communication.rs
@@ -13,11 +13,15 @@
 
 use bytes::Bytes;
 pub use notification::{NotificationError, NotificationListener, Notifier};
+use protobuf::Message;
 pub use pubsub::{PubSubError, Publisher, Subscriber};
 pub use rpc::{RpcClient, RpcServer, ServiceInvocationError};
 use std::{error::Error, fmt::Display};
 
-use crate::{UPayloadFormat, UPriority};
+use crate::{
+    umessage::{self, UMessageError},
+    UPayloadFormat, UPriority,
+};
 
 mod notification;
 mod pubsub;
@@ -54,12 +58,12 @@ impl Display for RegistrationError {
 
 impl Error for RegistrationError {}
 
-const DEFAULT_TTL: u16 = 10_000; // 10 seconds
+const DEFAULT_TTL: u32 = 10_000; // 10 seconds
 
 /// General options that clients might want to specify when sending a uProtocol message.
 #[derive(Debug)]
 pub struct CallOptions {
-    ttl: u16,
+    ttl: u32,
     token: Option<String>,
     priority: Option<UPriority>,
 }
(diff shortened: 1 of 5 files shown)


sophokles73 (approved):

sophokles73 (commented):
IMHO we will still need `extract_from_protobuf` functionality in UMessage because not all L2 API uses UPayload, e.g. the PubSub API uses UListener for handling messages ...

AnotherDaniel:
@sophokles73 alright this is banged into shape now - can you have a final look?

AnotherDaniel:
@sophokles73 alright this is banged into shape now - can you have a final look?

src/communication.rs
```diff
@@ -114,13 +181,64 @@ impl UPayload {
...
+    /// type to `UPayload::UPAYLOAD_FORMAT_PROTOBUF`.
+    ///
+    /// # Arguments
+    ///
+    /// * `message` - The protobuf `Message` to wrap in UPayload.
+    ///
+    /// # Returns
+    ///
+    /// `UPayload` with `Message` as `UPayloadType::UPAYLOAD_FORMAT_PROTOBUF`` payload.
+    pub fn from_protobuf<M>(message: M) -> Option<Self>
```
sophokles73, line 194: IMHO this should better return a Result with an error if the message cannot be serialized. However, we might also consider implementing `TryFrom<M: Message>` instead ...

src/communication/rpc.rs
```diff
@@ -91,6 +92,48 @@ pub trait RpcClient: Send + Sync {
     ) -> Result<Option<UPayload>, ServiceInvocationError>;
 }
 
+impl dyn RpcClient {
```
sophokles73, line 95: Shouldn't this simply be a default implementation in the trait instead?

src/umessage.rs
```diff
@@ -64,10 +66,37 @@ impl From<&str> for UMessageError {
 }
 
 impl UMessage {
-    /// Extracts the payload-contained protobuf message from a `UMessage`.
+    // Deserialize a proto-generated `Message`-type from payload `Bytes`, according to `UPayloadFormat`
+    // Will only succeed if payload format is one of
+    // - `UPayloadFormat::UPAYLOAD_FORMAT_PROTOBUF`
+    // - `UPayloadFormat::UPAYLOAD_FORMAT_PROTOBUF_WRAPPED_IN_ANY`
+    pub(crate) fn deserialize_protobuf_bytes<T: Message + Default>(
```
sophokles73, line 73: FMPOV we should move this out of the UMessage struct as it has meaning independently from it

sophokles73 (changes requested):

src/communication/rpc.rs
AnotherDaniel: Nope - those have to be statically resolvable at compile-time, which doesn't work with generic types.

src/communication.rs
AnotherDaniel: Can't make the TryFrom work, conflict with generic implemenation from core... went for using Result<>.

sophokles73 (approved):
LGTM

PR #129, opened 2024-06-26
Support adding traceparent to UMessage

### LINKED ISSUE #128: UMessageBuilder does not support setting traceparent
The [UAttributes spec](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uattributes.adoc#2-common-message-attributes) allows adding a W3C Trace Context to a UMessage's attributes. However, the UMessageBuilder currently does not support this.

sophokles73:
UMessageBuilder has been extended to support setting a W3C Trace
Context identifier to convey in a message.

Fixes #128

### DIFF
--- a/src/communication.rs
+++ b/src/communication.rs
@@ -20,7 +20,7 @@ use std::{error::Error, fmt::Display};
 
 use crate::{
     umessage::{self, UMessageError},
-    UPayloadFormat, UPriority,
+    UPayloadFormat, UPriority, UUID,
 };
 
 mod notification;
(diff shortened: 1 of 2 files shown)


AnotherDaniel (approved):
LGTM

AnotherDaniel:
Out of interest - quoting the W3C Spec: 

> At a minimum they MUST propagate the traceparent and tracestate headers and guarantee traces are not broken. This behavior is also referred to as forwarding a trace.

So shouldn't we also add a `tracestate` field?

PR #131, opened 2024-06-26
Add UTransport based RpcClient implementation

sophokles73:
Added an RpcClient implementation that keeps all state
about pending requests in memory only. Also added
corresponding unit tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,13 @@ once_cell = { version = "1.19" }
 protobuf = { version = "3.3", features = ["with-bytes"] }
 rand = { version = "0.8" }
 regex = { version = "1.10" }
+tokio = { version = "1.35", default-features = false, features = [
+    "macros",
+    "rt",
+    "rt-multi-thread",
+    "sync",
+    "time",
+] }
 uriparse = { version = "0.6" }
 uuid-simd = { version = "0.8", default-features = false, features = [
     "std",
@@ -59,8 +66,6 @@ protoc-bin-vendored = { version = "3.0" }
 reqwest = { version = "0.12", features = ["blocking"] }
 
 [dev-dependencies]
-async-std = { version = "1.12.0", features = ["attributes"] }
-futures = { version = "0.3.30" }
 test-case = { version = "3.3" }
 
 [profile.release]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,12 +47,17 @@ once_cell = { version = "1.19" }
 protobuf = { version = "3.3", features = ["with-bytes"] }
 rand = { version = "0.8" }
 regex = { version = "1.10" }
+thiserror = { version = "1.0.52" }
 tokio = { version = "1.35", default-features = false, features = [
     "macros",
     "rt",
     "rt-multi-thread",
     "sync",
     "time",
+] }
+tracing = { version = "0.1.35", default-features = false, features = [
+    "log",
+    "std",
 ] }
 uriparse = { version = "0.6" }
 uuid-simd = { version = "0.8", default-features = false, features = [
@@ -66,6 +71,7 @@ protoc-bin-vendored = { version = "3.0" }
 reqwest = { version = "0.12", features = ["blocking"] }
 
 [dev-dependencies]
+mockall = "0.12.1"
 test-case = { version = "3.3" }
 
 [profile.release]
--- a/src/uri.rs
+++ b/src/uri.rs
@@ -283,6 +283,17 @@ impl UUri {
         self.eq(&UUri::default())
     }
 
+    /// Gets a URI that consists of wildcards only and therefore matches any URI.
+    pub fn any() -> Self {
+        UUri {
+            authority_name: WILDCARD_AUTHORITY.to_string(),
+            ue_id: WILDCARD_ENTITY_ID,
+            ue_version_major: WILDCARD_ENTITY_VERSION,
+            resource_id: WILDCARD_RESOURCE_ID,
+            ..Default::default()
+        }
+    }
+
     /// Serializes this UUri to a URI string.
     ///
     /// # Arguments
--- a/src/utransport.rs
+++ b/src/utransport.rs
@@ -70,7 +70,7 @@ pub trait LocalUriProvider: Send + Sync {
 /// use up_rust::{UListener, UMessage, UStatus};
 ///
 /// use async_trait::async_trait;
-/// use async_std::task;
+/// use tokio::task;
 /// use std::sync::{Arc, Mutex};
 ///
 /// #[derive(Clone)]
@@ -84,7 +84,7 @@ pub trait LocalUriProvider: Send + Sync {
 /// #[async_trait]
 /// impl UListener for LongTaskListener {
 ///     async fn on_receive(&self, msg: UMessage) {
-///         task::spawn(send_to_jupiter(msg));
+///         tokio::spawn(send_to_jupiter(msg));
 ///     }
 /// }
 /// ```
@@ -347,7 +347,6 @@ mod tests {
     use crate::ComparableListener;
     use crate::UListener;
     use crate::{UCode, UMessage, UStatus, UTransport, UUri};
-    use async_std::task;
     use async_trait::async_trait;
     use std::collections::{HashMap, HashSet};
     use std::sync::Arc;
@@ -380,7 +379,7 @@ mod tests {
                         for listener in listeners.iter() {
                             let task_listener = listener.clone();
                             let task_umessage = umessage.clone();
-                            task::spawn(
+                            tokio::spawn(
                                 async move { task_listener.on_receive(task_umessage).await },
                             );
                         }
@@ -489,13 +488,14 @@ mod tests {
         }
     }
 
-    #[test]
-    fn test_register_and_receive() {
+    #[tokio::test]
+    async fn test_register_and_receive() {
         let mut up_client_foo = UPClientFoo::default();
         let uuri_1 = uuri_factory(1);
         let listener_baz: Arc<dyn UListener> = Arc::new(ListenerBaz);
-        let register_res =
-            task::block_on(up_client_foo.register_listener(&uuri_1, None, listener_baz.clone()));
+        let register_res = up_client_foo
+            .register_listener(&uuri_1, None, listener_baz.clone())
+            .await;
         assert!(register_res.is_ok());
 
         let umessage = UMessage::default();
@@ -503,21 +503,23 @@ mod tests {
         assert_eq!(check_on_receive_res, Ok(()));
     }
 
-    #[test]
-    fn test_register_and_unregister() {
+    #[tokio::test]
+    async fn test_register_and_unregister() {
         let mut up_client_foo = UPClientFoo::default();
         let uuri_1 = uuri_factory(1);
         let listener_baz: Arc<dyn UListener> = Arc::new(ListenerBaz);
-        let register_res =
-            task::block_on(up_client_foo.register_listener(&uuri_1, None, listener_baz.clone()));
+        let register_res = up_client_foo
+            .register_listener(&uuri_1, None, listener_baz.clone())
+            .await;
         assert!(register_res.is_ok());
 
         let umessage = UMessage::default();
         let check_on_receive_res = up_client_foo.check_on_receive(&uuri_1, &umessage);
         assert_eq!(check_on_receive_res, Ok(()));
 
-        let unregister_res =
-            task::block_on(up_client_foo.unregister_listener(&uuri_1, None, listener_baz));
+        let unregister_res = up_client_foo
+            .unregister_listener(&uuri_1, None, listener_baz)
+            .await;
         assert!(unregister_res.is_ok());
 
         let umessage = UMessage::default();
@@ -525,30 +527,34 @@ mod tests {
         assert!(check_on_receive_res.is_err());
     }
 
-    #[test]
-    fn test_register_multiple_listeners_on_one_uuri() {
+    #[tokio::test]
+    async fn test_register_multiple_listeners_on_one_uuri() {
         let mut up_client_foo = UPClientFoo::default();
         let uuri_1 = uuri_factory(1);
         let listener_baz: Arc<dyn UListener> = Arc::new(ListenerBaz);
         let listener_bar: Arc<dyn UListener> = Arc::new(ListenerBar);
 
-        let register_res =
-            task::block_on(up_client_foo.register_listener(&uuri_1, None, listener_baz.clone()));
+        let register_res = up_client_foo
+            .register_listener(&uuri_1, None, listener_baz.clone())
+            .await;
         assert!(register_res.is_ok());
 
-        let register_res =
-            task::block_on(up_client_foo.register_listener(&uuri_1, None, listener_bar.clone()));
+        let register_res = up_client_foo
+            .register_listener(&uuri_1, None, listener_bar.clone())
+            .await;
         assert!(register_res.is_ok());
 
         let umessage = UMessage::default();
         let check_on_receive_res = up_client_foo.check_on_receive(&uuri_1, &umessage);
         assert_eq!(check_on_receive_res, Ok(()));
 
-        let unregister_baz_res =
-            task::block_on(up_client_foo.unregister_listener(&uuri_1, None, listener_baz));
+        let unregister_baz_res = up_client_foo
+            .unregister_listener(&uuri_1, None, listener_baz)
+            .await;
         assert!(unregister_baz_res.is_ok());
-        let unregister_bar_res =
-            task::block_on(up_client_foo.unregister_listener(&uuri_1, None, listener_bar));
+        let unregister_bar_res = up_client_foo
+            .unregister_listener(&uuri_1, None, listener_bar)
+            .await;
         assert!(unregister_bar_res.is_ok());
 
         let check_on_receive_res = up_client_foo.check_on_receive(&uuri_1, &umessage);
@@ -566,41 +572,40 @@ mod tests {
         assert!(check_on_receive_res.is_err());
     }
 
-    #[test]
-    fn test_register_multiple_same_listeners_on_one_uuri() {
+    #[tokio::test]
+    async fn test_register_multiple_same_listeners_on_one_uuri() {
         let mut up_client_foo = UPClientFoo::default();
         let uuri_1 = uuri_factory(1);
 
         let listener_baz: Arc<dyn UListener> = Arc::new(ListenerBaz);
-        let register_res =
-            task::block_on(up_client_foo.register_listener(&uuri_1, None, listener_baz.clone()));
+        let register_res = up_client_foo
+            .register_listener(&uuri_1, None, listener_baz.clone())
+            .await;
         assert!(register_res.is_ok());
 
-        let register_res =
-            task::block_on(up_client_foo.register_listener(&uuri_1, None, listener_baz.clone()));
+        let register_res = up_client_foo
+            .register_listener(&uuri_1, None, listener_baz.clone())
+            .await;
         assert!(register_res.is_err());
 
         let listener_baz_completely_different: Arc<dyn UListener> = Arc::new(ListenerBaz);
-        let register_res = task::block_on(up_client_foo.register_listener(
-            &uuri_1,
-            None,
-            listener_baz_completely_different.clone(),
-        ));
+        let register_res = up_client_foo
+            .register_listener(&uuri_1, None, listener_baz_completely_different.clone())
+            .await;
         assert!(register_res.is_ok());
 
         let umessage = UMessage::default();
         let check_on_receive_res = up_client_foo.check_on_receive(&uuri_1, &umessage);
         assert_eq!(check_on_receive_res, Ok(()));
 
-        let unregister_res = task::block_on(up_client_foo.unregister_listener(
-            &uuri_1,
-            None,
-            listener_baz_completely_different,
-        ));
+        let unregister_res = up_client_foo
+            .unregister_listener(&uuri_1, None, listener_baz_completely_different)
+            .await;
         assert!(unregister_res.is_ok());
 
-        let unregister_baz_res =
-            task::block_on(up_client_foo.unregister_listener(&uuri_1, None, listener_baz));
+        let unregister_baz_res = up_client_foo
+            .unregister_listener(&uuri_1, None, listener_baz)
+            .await;
         assert!(unregister_baz_res.is_ok());
 
         let check_on_receive_res = up_client_foo.check_on_receive(&uuri_1, &umessage);
--- a/src/utransport.rs
+++ b/src/utransport.rs
@@ -61,10 +61,6 @@ pub trait LocalUriProvider: Send + Sync {
 ///             *inner_foo = format!("latest message length: {}", payload.len());
 ///         }
 ///     }
-///
-///     async fn on_error(&self, err: UStatus) {
-///         println!("uh oh, we got an error: {err:?}");
-///     }
 /// }
 /// ```
 ///
@@ -90,10 +86,6 @@ pub trait LocalUriProvider: Send + Sync {
 ///     async fn on_receive(&self, msg: UMessage) {
 ///         task::spawn(send_to_jupiter(msg));
 ///     }
-///
-///     async fn on_error(&self, err: UStatus) {
-///         println!("unable to send to jupiter :( {err:?}");
-///     }
 /// }
 /// ```
 #[async_trait]
@@ -115,24 +107,6 @@ pub trait UListener: Send + Sync {
     /// Because `on_receive()` is async you may choose to either `.await` it in the current context
     /// or spawn it onto a new task and await there to allow current context to immediately continue.
     async fn on_receive(&self, msg: UMessage);
-
-    /// Performs some action on receipt of an error.
-    ///
-    /// # Parameters
-    ///
-    /// * `err` - The error as `UStatus`
-    ///
-    /// # Note for `UListener` implementers
-    ///
-    /// `on_error()` is expected to return almost immediately. If it does not, it could potentially
-    /// block further message receipt. For long-running operations consider passing off received
-    /// error to a different async function to handle it and returning.
-    ///
-    /// # Note for `UTransport` implementers
-    ///
-    /// Because `on_error()` is async you may choose to either `.await` it in the current context
-    /// or spawn it onto a new task and await there to allow current context to immediately continue.
-    async fn on_error(&self, err: UStatus);
 }
 
 /// [`UTransport`] is the uP-L1 interface that provides a common API for uE developers to send and receive messages.
@@ -247,10 +221,6 @@ pub trait UTransport: Send + Sync {
     /// # #[async_trait]
     /// # impl UListener for MyListener {
     /// #     async fn on_receive(&self, msg: UMessage) {
-    /// #         todo!()
-    /// #     }
-    /// #
-    /// #     async fn on_error(&self, err: UStatus) {
     /// #         todo!()
     /// #     }
     /// # }
@@ -498,10 +468,6 @@ mod tests {
         async fn on_receive(&self, msg: UMessage) {
             println!("Printing msg from ListenerBaz! received: {:?}", msg);
         }
-
-        async fn on_error(&self, err: UStatus) {
-            println!("Printing err from ListenerBaz! received {:?}", err)
-        }
     }
 
     #[derive(Clone, Debug)]
@@ -510,10 +476,6 @@ mod tests {
     impl UListener for ListenerBar {
         async fn on_receive(&self, msg: UMessage) {
             println!("Printing msg from ListenerBar! received: {:?}", msg);
-        }
-
-        async fn on_error(&self, err: UStatus) {
-            println!("Printing err from ListenerBar! received: {:?}", err);
         }
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -42,6 +42,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstyle"
+version = "1.0.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "038dfcf04a5feb68e9c60b21c9625a54c2c0616e79b72b0fd87075a056ae1d1b"
+
+[[package]]
 name = "anyhow"
 version = "1.0.81"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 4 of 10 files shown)


src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,569 @@
...
+            .response_listener
+            .try_add_pending_request(message_id.clone())?;
+        self.transport
+            .send(rpc_request_message)
+            .await
+            .map_err(|e| {
+                self.response_listener.remove_pending_request(&message_id);
+                e
+            })?;
+        tokio::time::timeout(
```
evshary, line 279: Would it be easier if we put the response data into async queue according to its reqid in the listener, and then retrieve the data here?

Cargo.toml
```diff
@@ -47,6 +47,14 @@ once_cell = { version = "1.19" }
 protobuf = { version = "3.3", features = ["with-bytes"] }
 rand = { version = "0.8" }
 regex = { version = "1.10" }
+thiserror = { version = "1.0.52" }
+tokio = { version = "1.35", default-features = false, features = [
```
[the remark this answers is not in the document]
sophokles73: I agree

src/utransport.rs
```diff
@@ -115,24 +107,6 @@ pub trait UListener: Send + Sync {
...
-    ///
-    /// `on_error()` is expected to return almost immediately. If it does not, it could potentially
-    /// block further message receipt. For long-running operations consider passing off received
-    /// error to a different async function to handle it and returning.
-    ///
-    /// # Note for `UTransport` implementers
-    ///
-    /// Because `on_error()` is async you may choose to either `.await` it in the current context
-    /// or spawn it onto a new task and await there to allow current context to immediately continue.
-    async fn on_error(&self, err: UStatus);
```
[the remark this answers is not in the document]
sophokles73: invoke the on_message function with a UMessage representing the error. The UMessage contains all necessary context to find the corresponding request that the error is the outcome of.

src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,569 @@
...
+            {
+                self.pending_request_store
+                    .complete_response_future(&reqid, msg)
+            }
+        }
+    }
+}
+
+/// An ['RpcClient'] which keeps all information about pending requests in memory.
+pub struct InMemoryRpcClient {
```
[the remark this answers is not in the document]
sophokles73: In large scale (cloud) environments you could think of keeping the state in a shared persistent store to provide for horizontal scalability. Not sure, though, if this will ever be part of the up-rust crate. But FMPOV it still makes sense to make the implementation strategy transparent so that users know what they can expect.

src/communication/rpc_impl.rs
sophokles73: Not sure I can follow. Can you elaborate a little?

src/utransport.rs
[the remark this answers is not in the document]
PLeVasseur: I remember this coming up as a question for this API when discussing uP-L2 a while back.

I liked your concept originally with UListener of having two explicit paths with on_receive and on_error.

Could you sketch out in this PR what it would look like doing handling of the error returned on a UListener for the four messages types when using a UTransport: Publish, Notification, Request, Response?

If I've missed it, please point it out :slightly_smiling_face:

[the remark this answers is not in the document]
sophokles73: I don't think I can follow. Can you be a little more specific regarding what is unclear?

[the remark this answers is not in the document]
evshary: > invoke the on_message function with a UMessage representing the error. The UMessage contains all necessary context to find the corresponding request that the error is the outcome of.

I guess you mean `on_receive`? To be honest, I prefer to handle the errors in another function to make the code clean, but it's just my preference.

src/communication/rpc_impl.rs
evshary: I'm not sure whether I missed anything, but I'm considering the possibility of avoiding implementing the Future. Here are my thoughts.

```rust
async fn invoke_method(...) {
     // Get (tx, rx) from async_channel
     // Put (reqid, tx) into HashMap
     // Waiting rx with tokio timeout
}

async fn on_receive(...) {
    // Use reqid to get tx in the HashMap
    // Put UMessage into tx
}
```

src/utransport.rs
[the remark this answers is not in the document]
PLeVasseur: Yes, @evshary is correct. What I mean @sophokles73 to ask about is how errors would be identified and handled by the UListener. My preference would be to have an explicitly different path for the handling of those errors with on_error().

Could you help me by showing examples of handling the error path within the UListener with the single on_receive()?

src/communication/rpc_impl.rs
sophokles73: Sounds like a great idea, let me try if I get this running ...

src/utransport.rs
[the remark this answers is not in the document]
sophokles73: In the [uProtocol Error Model](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/error_model.adoc#2-service-implementation) I tried to establish the pattern that a `commstatus` that is not `UCode::OK` indicates an error and that the receiver of the message should assume that the payload contains a protobuf encoded `UStatus` with error details.

However, some existing service implementations still return a message with a `commstatus` of `UCode::OK` but then include an (error) `UStatus` in the operations standard response message. So only the client (invoker) of the operation will ultimately know how an error manifests itself in the response message.

A uEntity will therefore (still) need to know about the error model implemented by the service operations that it invokes. The RpcClient already implements the strategy outlined above and tries to map a UMessage with a `commstatus != UCode::OK` to a `ServiceInvocationError`.

For a UTransport implementation, I would suggest to create a UMessage following the recommended error model if it runs into an unresolvable error situation and needs to inform upper layers about it. In all other cases, it should simply invoke `UListener::on_receive` with whatever message it has received from the underlying messaging infrastructure.

src/communication/rpc_impl.rs
sophokles73: I got this up and running, @evshary. I will update the PR soon ...

sophokles73: would you mind taking another look @evshary ?

evshary (approved):
LGTM for the changes

sophokles73:
@AnotherDaniel any thoughts/concerns?

src/utransport.rs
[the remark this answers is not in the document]
PLeVasseur: Hey @sophokles73 -- reading through this now. Took last Friday off.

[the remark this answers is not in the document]
PLeVasseur: Thanks for the link to the spec! Nicely written.

Now that I see the spec, this makes more sense. So within a UListener rather than having two explicit flows, one for error and one for successful execution, the implementer of a UListener would instead check the commstatus if they were specifically listening for Responses as in a Service Consumer style uEntity.

Is it correct to say that as [outlined in the UAttributes spec](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uattributes.adoc) that the commstatus only applies to Response messages?

That in the case of Notification, Publish, and Request messages that if these messages when attempting to be sent would be in some erroneous state with commstatus set to something or something not OK they simply would not be transmitted at all by the UTransport?

Therefore that a UListener which is pinged that has been registered for Notification, Publish, or Request style messages should assume that the message it received is not in error?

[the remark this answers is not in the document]
sophokles73: A client trying to send a Request message using UTransport:.send would get back an Err(UStatus) if the message cannot be sent. The receiving side (service provider) UListener in this case would never even get the Request message. How could it, if sending has failed?

For the same reason, a UListener listening for Notifications will simply not receive any messages for send attempts that have failed on the client (sender) side ...

src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,486 @@
...
+    /// * `transport` - The uProtocol Transport Layer implementation to use for invoking service operations.
+    /// * `uri_provider` - The helper for creating URIs that represent local resources.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the generic RPC Response listener could not be
+    /// registered with the given transport.
+    pub async fn new(
+        transport: Arc<dyn UTransport>,
+        uri_provider: Arc<dyn LocalUriProvider>,
```
PLeVasseur, line 147: IIRC we left it open that e.g. UPClientZenoh could implement LocalUriProvider or impl this on a separate struct, right?

So in the former case this would look like:

```rust
let rpc_client = InMemoryRpcClient::new(up_client_zenoh.clone(), up_client_zenoh.clone());
```

right?

Are you thinking to add some recommendation on which path is better in the future as we learn more?

src/utransport.rs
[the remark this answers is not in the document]
sophokles73: > Therefore that a UListener which is pinged that has been registered for Notification, Publish, or Request style messages should assume that the message it received is not in error?

Yes, albeit it should, of course, check whether the UAttributes are consistent and that the payload can be processed ...

src/uri.rs
```diff
@@ -283,6 +283,17 @@ impl UUri {
         self.eq(&UUri::default())
     }
 
+    /// Gets a URI that consists of wildcards only and therefore matches any URI.
+    pub fn any() -> Self {
+        UUri {
```
PLeVasseur, line 288: :+1: 

Reminds me I need to open a PR to add a

```rust
pub fn any_with_authority_name(authority_name: &str) -> Self {
  authority_name: authority_name.to_string(),
  ue_id: WILDCARD_ENTITY_ID,
  ue_version_major: WILDCARD_ENTITY_VERSION,
  resource_id: WILDCARD_RESOURCE_ID,
  ..Default::default()
}
```

AnotherDaniel (approved):
LGTM - I'm sure I will run into some bumps when implementing based on this, but that's for future Daniel...

src/communication/rpc_impl.rs
sophokles73: This is correct. I am currently not in a position to recommend one or the other, so I'd rather let users decide what they prefer/see fit.

src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,486 @@
+/********************************************************************************
```
PLeVasseur, line 1: What do you think the chances are that we'd have > 1 RPC implementation in the future in up-rust?

Worthwhile to name this file something more specific s.t. in the future we can have `rpc_in_memory_impl.rs`, `rpc_db_backed_impl.rs`, and so on?

Or just leave this and we'll navigate that as it comes up?

src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,486 @@
...
+    fn contains_pending_request(&self, reqid: &UUID) -> bool {
+        self.response_listener.contains(reqid)
+    }
+}
+
+#[async_trait]
+impl RpcClient for InMemoryRpcClient {
+    async fn invoke_method(
+        &self,
+        method: UUri,
```
PLeVasseur, line 177: I'm not buying or selling here.

Seems like the [up-cpp](https://github.com/eclipse-uprotocol/up-cpp/blob/3b80662d884b189d36114ef2922a92bc3699a397/include/up-cpp/communication/RpcClient.h#L51) version bakes their method into the construction of their RpcClient vs here we do not and thus are able to use a single RpcClient for multiple different service method invocation types.

Wondering if you had some thoughts on the pros/cons.

Tagging @gregmedd for any thoughts as well.

src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,486 @@
...
+        call_options: CallOptions,
+        payload: Option<UPayload>,
+    ) -> Result<Option<UPayload>, ServiceInvocationError> {
+        let message_id = call_options.message_id().unwrap_or_else(UUID::build);
+
+        let mut builder = UMessageBuilder::request(
```
PLeVasseur, line 183: Not sure if this is expensive or not, but in the 99% case, we're probably calling just a handful of different methods and thus could probably cache these builders and modify just the CallOptions portions as necessary upon each call to `invoke_method()`.

Again, not trying to nudge any particular action, just an observation.

src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,486 @@
...
+        .map_err(|e| ServiceInvocationError::InvalidArgument(e.to_string()))?;
+
+        let receiver = self
+            .response_listener
+            .try_add_pending_request(message_id.clone())?;
+        self.transport
+            .send(rpc_request_message)
+            .await
+            .map_err(|e| {
+                self.response_listener.remove_pending_request(&message_id);
```
PLeVasseur, line 213: Very clean. I like this usage.

src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,486 @@
...
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+use std::collections::hash_map::Entry;
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
```
PLeVasseur, line 16: Have you considered using an async-aware primitive here instead for `Mutex`?

src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,486 @@
...
+        Err(ServiceInvocationError::RpcError(UStatus::fail_with_code(
+            UCode::INTERNAL,
+            "response message does not contain attributes",
+        )))
+    }
+}
+
+struct ResponseListener {
+    // request ID -> sender for response message
+    pending_requests: Mutex<HashMap<UUID, Sender<UMessage>>>,
```
PLeVasseur, line 58: I suppose the distribution of reads and writes is pretty balanced here and entries in `pending_requests` have fairly short stays.

May not be worthwhile to consider something like a `RwLock`.

src/communication/rpc_impl.rs
sophokles73: Good point, will rename the file :+1:

src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,486 @@
...
+}
+
+impl ResponseListener {
+    fn try_add_pending_request(
+        &self,
+        reqid: UUID,
+    ) -> Result<Receiver<UMessage>, ServiceInvocationError> {
+        if let Ok(mut pending_requests) = self.pending_requests.lock() {
```
PLeVasseur, line 66: You could save yourself some indentation here and do early returns by:

```rust
let Ok(mut pending_requests) = self.pending_requests.lock() else {
  return Err(ServiceInvocationError::RpcError(UStatus::fail_with_code(
      UCode::INTERNAL,
      "failed to add response handler",
  )));
};

let Entry::Vacant(entry) = pending_requests.entry(reqid) else {
  return Err(ServiceInvocationError::AlreadyExists(
      "RPC request with given ID already pending".to_string(),
  ));
};

let (tx, rx) = tokio::sync::oneshot::channel();
entry.insert(tx);
Ok(rx)
```

Just a though.

src/communication/rpc_impl.rs
sophokles73: I guess the most expensive part is serialization of the (per request) payload. Compared to that, I would assume that the costs for creating the builder itself and setting the other properties are neglectable. WDYT?

src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,486 @@
...
+        } else {
+            Err(ServiceInvocationError::RpcError(UStatus::fail_with_code(
+                UCode::INTERNAL,
+                "failed to add response handler",
+            )))
+        }
+    }
+
+    fn handle_response(&self, reqid: &UUID, response_message: UMessage) {
+        if let Ok(mut pending_requests) = self.pending_requests.lock() {
```
PLeVasseur, line 85: Same concept here about avoiding indentation and preferring early return. May just be a style thing? For your consideration:

```rust
let mut pending_requests = self.pending_requests.lock() else {
  // TODO: May be good to explicitly log some sort of error here stating that the Mutex has been poisoned?
  // Reference: https://doc.rust-lang.org/std/sync/type.LockResult.html
  return;
}

let Some(sender) = pending_requests.remove(reqid)  else {
  // we seem to have received a duplicate of the response message,
  // ignoring it
  // TODO: May be good to explicitly log that we already handled the previous duplicate response message?
  return;
}

let _ = sender.send(response_message);
```

src/communication/rpc_impl.rs
sophokles73: No, as there is no need to maintain a lock across an await point, I think we can/should stick with the standard Mutex. As the tokio docs point out:
```rust
/// # Which kind of mutex should you use?
///
/// Contrary to popular belief, it is ok and often preferred to use the ordinary
/// [`Mutex`][std] from the standard library in asynchronous code.
///
/// The feature that the async mutex offers over the blocking mutex is the
/// ability to keep it locked across an `.await` point. This makes the async
/// mutex more expensive than the blocking mutex, so the blocking mutex should
/// be preferred in the cases where it can be used. The primary use case for the
/// async mutex is to provide shared mutable access to IO resources such as a
/// database connection. If the value behind the mutex is just data, it's
/// usually appropriate to use a blocking mutex such as the one in the standard
/// library or [`parking_lot`].
```

src/communication/rpc_impl.rs
sophokles73: I guess not. All reads are done with the intention to remove the entry as well ...

src/communication/rpc_impl.rs
sophokles73: Actually, I find that much less readable ;-)

src/communication/rpc_impl.rs
sophokles73: what level would you log this at, DEBUG?

src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,486 @@
...
+            .map_or(false, |pending_requests| {
+                pending_requests.contains_key(reqid)
+            })
+    }
+}
+
+#[async_trait]
+impl UListener for ResponseListener {
+    async fn on_receive(&self, msg: UMessage) {
+        if msg.is_response() {
```
PLeVasseur, line 114: Sorry to harp on this, but again, I like reducing indentation / nesting :slightly_smiling_face: 

```rust
if !msg.is_response() {
  // TODO: Consider logging some sort of trace or info level message?
  return;
}

let Some(reqid) = msg
  .attributes
  .as_ref()
  .and_then(|attribs| attribs.reqid.clone().into_options()) else 
{
  // TODO: Consider logging some sort of trace or info level message?
  return;
}

self.handle_response(&reqid, msg);
```

src/communication/rpc_impl.rs
sophokles73: Not really but I guess when baking it into the constructor, there is much less dynamic memory allocation required during runtime, which might be desirable? Assuming that a client uEntity will have a fixed set of other uEntities it needs to interact with, having dedicated RpcClient instances for them should also be ok.

src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,486 @@
...
+
+#[async_trait]
+impl RpcClient for InMemoryRpcClient {
+    async fn invoke_method(
+        &self,
+        method: UUri,
+        call_options: CallOptions,
+        payload: Option<UPayload>,
+    ) -> Result<Option<UPayload>, ServiceInvocationError> {
+        let message_id = call_options.message_id().unwrap_or_else(UUID::build);
```
PLeVasseur, line 181: Might just be me, but I think it'd increase cohesion if we moved this line to right above where we call
```rust
builder.with_message_id(message_id.clone());
```

src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,486 @@
...
+            .response_listener
+            .try_add_pending_request(message_id.clone())?;
+        self.transport
+            .send(rpc_request_message)
+            .await
+            .map_err(|e| {
+                self.response_listener.remove_pending_request(&message_id);
+                e
+            })?;
+        match tokio::time::timeout(Duration::from_millis(call_options.ttl() as u64), receiver).await
```
PLeVasseur, line 216: Given we don't use the error path's content consider:

```rust
if let Ok(Ok(response_message)) = tokio::time::timeout(Duration::from_millis(call_options.ttl() as u64), receiver).await {
  handle_response_message(response_message)
} else {
  self.response_listener.remove_pending_request(&message_id);
  Err(ServiceInvocationError::DeadlineExceeded)
}
```

Makes the line a little longer, so may then be useful to break up the latter half across a few lines, but presented for your consideration :slightly_smiling_face:

src/communication/rpc_impl.rs
sophokles73: FMPOV it is at the right position because the variable is also needed/used at the end of the function and hiding it within the creation of the builder makes it harder to spot.

src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,486 @@
...
+};
+
+use super::{CallOptions, RpcClient, ServiceInvocationError, UPayload};
+
+fn handle_response_message(response: UMessage) -> Result<Option<UPayload>, ServiceInvocationError> {
+    if let Some(attribs) = response.attributes.as_ref() {
+        match attribs.commstatus.map(|v| v.enum_value_or_default()) {
+            Some(UCode::OK) | None => {
+                // successful invocation
+                response.payload.map_or(Ok(None), |payload| {
```
PLeVasseur, line 33: May be more of a spec question, but given that I think you wrote or rewrote this part of the spec, I'm wondering why it's [considered non-mandatory](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uattributes.adoc#6-rpc-response-message-attributes) for a Response message to not contain a commstatus.

Seems like to me that's a fairly reasonable thing to require.

src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,486 @@
...
+use crate::{
+    LocalUriProvider, UCode, UListener, UMessage, UMessageBuilder, UStatus, UTransport, UUri, UUID,
+};
+
+use super::{CallOptions, RpcClient, ServiceInvocationError, UPayload};
+
+fn handle_response_message(response: UMessage) -> Result<Option<UPayload>, ServiceInvocationError> {
+    if let Some(attribs) = response.attributes.as_ref() {
+        match attribs.commstatus.map(|v| v.enum_value_or_default()) {
+            Some(UCode::OK) | None => {
```
PLeVasseur, line 31: May be more of a spec question, but given that I think you wrote or rewrote this part of the spec, I'm wondering why it's [considered non-mandatory](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uattributes.adoc#6-rpc-response-message-attributes) for a Response message to not contain a commstatus.

Seems like to me that's a fairly reasonable thing to require.

src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,486 @@
...
+            Some(code) => {
+                // try to extract UStatus from response payload
+                let status = response.extract_protobuf().unwrap_or_else(|_e| {
+                    UStatus::fail_with_code(code, "failed to invoke service operation")
+                });
+                Err(ServiceInvocationError::from(status))
+            }
+        }
+    } else {
+        Err(ServiceInvocationError::RpcError(UStatus::fail_with_code(
```
PLeVasseur, line 49: Another suggestion here for early-return style by using `let Some(attribs) = ...` above.

src/communication/rpc_impl.rs
sophokles73: By default, the commstatus is assumed to be OK. No need to transmit it. Similarly, the default Priority for Requests is CS4, if not explicitly specified.

src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,486 @@
...
+                        attribs.id.as_ref() == Some(&expected_message_id)
+                            && attribs.priority.value() == UPriority::UPRIORITY_CS6.value()
+                            && attribs.ttl == Some(5_000)
+                            && attribs.token == Some("my_token".to_string())
+                            && attribs.traceparent == Some(traceparent.to_string())
+                    })
+            })
+            .returning(move |request_message| {
+                let request_payload: StringValue = request_message.extract_protobuf().unwrap();
+                let response_payload = StringValue {
```
PLeVasseur, line 374: Neat, didn't know about StringValue. That's useful, I can probably clean up some example code :slightly_smiling_face:

src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,486 @@
...
+
+        let mut mock_transport = MockTransport::default();
+        mock_transport.expect_do_register_listener().returning(
+            move |_source_filter, _sink_filter, listener| {
+                let mut v = listener_ref.lock().unwrap();
+                *v = Some(listener);
+                Ok(())
+            },
+        );
+        // and a remote service oration that returns an error
```
PLeVasseur, line 424:
```suggestion
        // and a remote service operation that returns an error
```

?

src/communication/rpc_impl.rs
```diff
@@ -0,0 +1,486 @@
...
+    async fn test_invoke_method_times_out() {
+        // GIVEN an RPC client
+        let mut mock_transport = MockTransport::default();
+        mock_transport
+            .expect_do_register_listener()
+            .returning(move |_source_filter, _sink_filter, _listener| Ok(()));
+        // and a remote service operation that does not return a response
+        mock_transport
+            .expect_do_send()
+            .returning(move |_request_message| Ok(()));
```
PLeVasseur, line 467: Interesting. So in this case your mock transport nominally would succeed if not for the very small TTL of 20ms?

Am I reading this correctly?

src/communication/rpc_impl.rs
PLeVasseur: Yeah, that seems reasonable to me.

src/communication/rpc_impl.rs
PLeVasseur: ;^}

src/communication/rpc_impl.rs
PLeVasseur: I agree with your characterization on costs. Seems fine to me for now, pending any kind of data to making the builder being some sort of hot spot.

src/communication/rpc_impl.rs
PLeVasseur: Yeah, frankly I think there's no one true right way here.

In the case of having it in the constructor, I think this fits the use-case pretty nicely of 99% of users of `RpcClient` and as you said _may_ reduce dynamic memory allocation.

On the other hand, having the _option_ for a single `RpcClient` is not a bad thing.

I suppose we could have some examples in here or in a future PR showing both styles to inspire users to the direction that makes sense to them.

Looking to see if @gregmedd has any thoughts here given that I now know that up-cpp is the outlier :^)

src/communication/rpc_impl.rs
sophokles73: The client side transport sends out the request successfully, but there is no response message being received. So the request will time out, regardless of the concrete TTL value. I used a very small one, because I do not want to delay the execution of the test case unnecessarily ...

PLeVasseur (approved):
Will nominally approve this one.

Would be good to see some logging added [here](https://github.com/eclipse-uprotocol/up-rust/pull/131#discussion_r1661090572)

src/communication/rpc_impl.rs
gregmedd: The two approaches are different philosophies. In `up-cpp`, we have applied the philosophy that one instance of a layer 2 object represents one association - this has been applied evenly across all of our layer 2 APIs. Specifically speaking to `RpcClient`, we expect the usage pattern to look something like:

```C++
class SomeRemoteEntity {
    ...
    RpcClient resource_foo;
    RpcClient resource_bar;
    ...
public:
    SomeRemoteEntity(/* options */) : resource_foo(/* setup foo */), resource_bar(/*setup bar*/) {}

    bool foo(/* some list of parameters */) {
        PayloadBuilder parameters;
        // Prepare parameters as payload
        ...
        auto future = resource_foo.invokeMethod(parameters);
        // Evaluate return value of future
        ...
        return foo_result;
    }
    ...
};
```

The setup for the RPC method occurs once, gets wrapped by some appropriate higher layer, then is reused for the life of the client entity. With that usage pattern, it make sense to have one `RpcClient` instance represent the association with one remote `RpcServer`'s method. In the example above, most code within a uE would call the `foo()` method on `SomeRemoteEntity` instead of using `RpcClient::invokeMethod()` directly since additional work is needed to prepare a request and process a response. This also has some other advantages, such as being able to track which pending requests belong to which `RpcClient`, allowing us to validate parameters earlier (i.e. during startup) rather than later (i.e. at the moment the method is invoked), and making the point of ownership for the instance more clear.

The alternative is to instead only require (at least) one `RpcClient` instance for a uE - a singleton-like `RpcClient` - where each call to `RpcClient::invokeMethod()` can invoke any possible remote method. To wrap those methods like the example above, the parameters needed to invoke the remote RPC method would need to be provided at each point of use. This makes it easier to reuse resources across the uses of `RpcClient` (such as a callback registration), with one of the tradeoffs being that validation is deferred to invocation.

In the `up-cpp` implementation of `RpcClient`, I plan to handle the sharing of some resources inside the `RpcClient` itself. The author of the uE doesn't need to worry about that detail - they can just create and hold `RpcClient` instances for RPC methods they want to invoke.

PLeVasseur: > In the example above, most code within a uE would call the `foo()` method on `SomeRemoteEntity` instead of using `RpcClient::invokeMethod()` directly since additional work is needed to prepare a request and process a response.

This also seems to be more amenable to some of the code generation tools that @stevenhartley has discussed, perhaps? WDYT Steven?

In addition, I really like this:

> allowing us to validate parameters earlier (i.e. during startup) rather than later (i.e. at the moment the method is invoked)

sophokles73: On the other hand it also means that you have to use the same parameters for each invocation, e.g. the same TTL, priority, token etc. Not sure if this flexibility is needed/helpful, though.

I am fine with both approaches.
@PLeVasseur @AnotherDaniel do you have any preference?

sophokles73:
@PLeVasseur I have added some log statements, does that work for you?

PLeVasseur:
> @PLeVasseur I have added some log statements, does that work for you?

Logging looks good to me. Thanks!

src/communication/rpc_impl.rs
PLeVasseur: I like the idea of supporting the 99% use case a little more obviously and front-loading validation upon construction, so I'm leaning towards the approach of having the service method, ttl, etc supplied upon construction.

Anyone else have thoughts? @evshary?

AnotherDaniel: I actually prefer the more flexible method, where the parameters are not front-loaded. As per current experience in the usubscription code base that I'm working on - there are some places where this would be a major nuisance, as I need dynamic parameters (e.g. target UUris) every time.

So if we do a front-load version, I strongly suggest that this becomes a wrapper around the more generic one, and both are available for use (which probably would result in another source of confusion though...).

evshary: I don't have strong opinions on this. It depends on how users use it. If users need to send requests to several methods, the dynamic one makes more sense. On the other hand, if users only stick to the same RPC Server, it's better to keep the front-loaded parameters. I might agree with @AnotherDaniel, since he is the user now :laughing:

sophokles73: The let's merge this for now and the other variant later, if we have a clear picture of when it would be more advantageous.

PR #132, opened 2024-06-26
comm_status() now takes a UCode

### LINKED ISSUE #130: UMessageBuilder.with_comm_status() should use UCode as parameter value (instead of i32)
Currently, UMessageBuilder.with_comm_status() is defined as follows:

    `pub fn with_comm_status(&mut self, comm_status: i32) -> &mut UMessageBuilder;`

The UMessageBuilder comm_status member-type is UCode - and so the only thing that this method signature does is, it forces users to put in `UCode::OK.value()` (requiring the non-obvious extra import of `protobuf::Enum`), just so that with_comm_status() then directly translates the i32 back into a UCode.

So I propose to change this sig to just use `comm_status: UCode`. This is not going to take away any flexibility from us, as we are (these days) owning and maintaining our own definition of UCode (not using the outside Google ones any more)

Any objections @sophokles73 @evshary @PLeVasseur @stevenhartley ?

AnotherDaniel:
Fix #130

### DIFF
--- a/src/umessage/umessagebuilder.rs
+++ b/src/umessage/umessagebuilder.rs
@@ -502,26 +502,24 @@ impl UMessageBuilder {
     /// # Examples
     ///
     /// ```rust
-    /// use protobuf::{Enum, EnumOrUnknown};
     /// use up_rust::{UCode, UMessageBuilder, UPriority, UUID, UUri};
     ///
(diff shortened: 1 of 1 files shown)


sophokles73 (approved):
LGTM

PR #136, opened 2024-07-04
Make CallOptions immutable

sophokles73:
Changed CallOptions struct to only support setting fields by means
of constructors, effectively making it immutable. This way
we will have better control over which fields may or may not be set
for the different types of messages to use the CallOptions for.

Also removed the traceparent field because it should not be necessary
to explicitly set it. Instead, when a UMessage is created, the current
span will be looked up in the current thread/future context.

This is for #122

### DIFF
--- a/src/communication.rs
+++ b/src/communication.rs
@@ -60,8 +60,6 @@ impl Display for RegistrationError {
 
 impl Error for RegistrationError {}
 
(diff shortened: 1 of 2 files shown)


AnotherDaniel (approved):
LGTM

PR #139, opened 2024-07-08
Expose the useful consts in UUri

evshary:
Since `up-transport-xxx-rust` needs to use the common WILDCARD const to process the UUri, it's useful to expose it from up-rust.


sophokles73 (approved):
LGTM

PR #140, opened 2024-07-08
Add UTransport based L2 Notification API implementation

sophokles73:
This is for #122

### DIFF
--- a/src/communication/default_notifier.rs
+++ b/src/communication/default_notifier.rs
@@ -60,36 +60,28 @@ impl Notifier for SimpleNotifier {
 
     async fn start_listening(
         &self,
-        origin_filter: &UUri,
+        topic: &UUri,
         listener: Arc<dyn UListener>,
     ) -> Result<(), RegistrationError> {
-        origin_filter
+        topic
             .verify_no_wildcards()
             .map_err(|e| RegistrationError::InvalidFilter(e.to_string()))?;
         self.transport
-            .register_listener(
-                origin_filter,
-                Some(&self.uri_provider.get_source_uri()),
-                listener,
-            )
+            .register_listener(topic, Some(&self.uri_provider.get_source_uri()), listener)
             .await
             .map_err(RegistrationError::from)
     }
 
     async fn stop_listening(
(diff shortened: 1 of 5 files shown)


AnotherDaniel (approved):
LGTM

src/communication/default_notifier.rs
```diff
@@ -0,0 +1,338 @@
...
+            .map_err(|e| NotificationError::InvalidArgument(e.to_string()))?;
+        self.transport
+            .send(msg)
+            .await
+            .map_err(NotificationError::NotifyError)
+    }
+
+    async fn start_listening(
+        &self,
+        origin_filter: &UUri,
```
evshary, line 63: We're using `topic` in the trait. Perhaps we can have a consistent name?

src/communication/default_notifier.rs
```diff
@@ -0,0 +1,338 @@
...
+    pub fn new(transport: Arc<dyn UTransport>, uri_provider: Arc<dyn LocalUriProvider>) -> Self {
+        SimpleNotifier {
+            transport,
+            uri_provider,
+        }
+    }
+}
+
+#[async_trait]
+impl Notifier for SimpleNotifier {
```
evshary, line 40: Should we separate `notify` and `start_listening` & `stop_listening` to different traits? If I understand correctly, `notify` is only used by the sender, while the other two are used by the receiver.

evshary (approved):
LGTM overall. I left some comments

src/communication/default_notifier.rs
sophokles73: I have taken inspiration from the [up-java implementation](https://github.com/eclipse-uprotocol/up-java/blob/main/src/main/java/org/eclipse/uprotocol/communication/Notifier.java) where everything is in a single interface. Do you think it has a lot of value to have separate traits here?

evshary: Hmm... Having different traits is clearer, but I prefer to have a more consistent API within different languages. So, it makes sense to me to have the same way as up-java does.

PR #141, opened 2024-07-09
Update MSRV to 1.74

sophokles73:
Increased the Minimum Supported Rust Version to accomodate for
requirements of the protobuf crate when using MessageFull instead of
the standard Message trait.

This should fix the nightly build which currently fails due to the
MSRV check.


evshary (approved):
LGTM

PR #144, opened 2024-07-11
Update NOTICE, remove 3rd party dep info

AnotherDaniel:


stevenhartley:
@AnotherDaniel why did you remove this?

AnotherDaniel:
Because it's no longer true - we're not pulling in the google .proto definition anymore.

PR #145, opened 2024-07-12
Add automock attributes to commonly used traits

sophokles73:
Added support for easily creating mock instances of commonly used
traits. This should make implementing unit tests involving these
types much easier.


AnotherDaniel (approved):
LGTM - more things to catch up with...

PR #146, opened 2024-07-12
Add RpcClient based implementation of USubscription

sophokles73:
Added an implementation of the USubscription trait that works
with any RpcClient implementation.

Also added the "communication" feature which is enabled by default
and provides the Communication Layer API types and implementation.
The "communication" feature includes the "usubscription" feature.

This is for #122

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,10 +33,11 @@ rust-version = "1.74.1"
 version = "0.1.5"
 
 [features]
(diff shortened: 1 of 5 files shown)


AnotherDaniel (approved):
Looks good! Another todo off my list, without even having to ask... ;-)

evshary (approved):
LGTM

PR #147, opened 2024-07-15
Add L2 Publisher implementation

sophokles73:
Added an implementation of the Communication Layer Publisher and
Subscriber traits which use the Transport Layer and implementations
of other Communication Layer clients.

This is for last client implementation missing for #122.

### DIFF
--- a/src/communication/default_pubsub.rs
+++ b/src/communication/default_pubsub.rs
@@ -1,0 +1,1188 @@
+/********************************************************************************
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+use std::{
+    collections::{hash_map::Entry, HashMap},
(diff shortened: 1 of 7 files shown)


src/communication/default_pubsub.rs
```diff
@@ -0,0 +1,1188 @@
...
+    ) -> Result<Self, RegistrationError> {
+        let rpc_client = InMemoryRpcClient::new(transport.clone(), uri_provider.clone())
+            .await
+            .map(Arc::new)?;
+        let usubscription_client = Arc::new(RpcClientUSubscription::new(rpc_client));
+        let notifier = Arc::new(SimpleNotifier::new(transport.clone(), uri_provider.clone()));
+        Self::for_clients(transport, uri_provider, usubscription_client, notifier).await
+    }
+
+    /// Creates a new client for given clients
```
AnotherDaniel, line 256: This would benefit from rewording/being a bit more verbose

src/communication/default_pubsub.rs
```diff
@@ -0,0 +1,1188 @@
...
+        Ok(InMemorySubscriber {
+            transport,
+            uri_provider,
+            usubscription,
+            notifier,
+            subscription_change_listener,
+        })
+    }
+
+    pub async fn stop(&self) -> Result<(), RegistrationError> {
```
AnotherDaniel, line 284: Public function should get a docstring, even if it's a short one

AnotherDaniel (approved):
LGTM - minor suggestions around documentation

AnotherDaniel (approved):
LGTM

PR #149, opened 2024-07-16
Include the const exposure in lib.rs

evshary:
As a following PR of https://github.com/eclipse-uprotocol/up-rust/pull/139. I forgot to include the exposure in `lib.rs`...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -58,11 +58,8 @@ pub use uattributes::{UAttributes, UAttributesError, UMessageType, UPayloadForma
 mod umessage;
 pub use umessage::{UMessage, UMessageBuilder, UMessageError};
 
-mod uri;
-pub use uri::{
-    UUri, UUriError, RESOURCE_ID_MIN_EVENT, RESOURCE_ID_RESPONSE, WILDCARD_AUTHORITY,
-    WILDCARD_ENTITY_ID, WILDCARD_ENTITY_VERSION, WILDCARD_RESOURCE_ID,
-};
+pub mod uri;
+pub use uri::{UUri, UUriError};
 
 mod ustatus;
 pub use ustatus::{UCode, UStatus};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,7 +59,10 @@ mod umessage;
 pub use umessage::{UMessage, UMessageBuilder, UMessageError};
 
 mod uri;
-pub use uri::{UUri, UUriError};
+pub use uri::{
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -59,7 +59,10 @@ mod umessage;
 pub use umessage::{UMessage, UMessageBuilder, UMessageError};
 
 mod uri;
-pub use uri::{UUri, UUriError};
+pub use uri::{
+    UUri, UUriError, RESOURCE_ID_MIN_EVENT, RESOURCE_ID_RESPONSE, WILDCARD_AUTHORITY,
```
sophokles73, line 63: hmm, I see the problem with the `uri` module not being public. However, I do not really like to expose these constants on the top level because then they are missing any context: it is quite unclear what `up_rust::RESOURCE_ID_MIN_EVENT` is useful for, isn't it? Maybe it would be better to make the `uri` module public instead?

evshary: Agree with you

AnotherDaniel:
Ok with me!

evshary:
> Ok with me! Only question is, do we really need to expose every symbol/const? I've really no idea, honest question.

At least for the UUri one, it's necessary for the transport layer (Zenoh, MQTT...). For others const, it depends. Maybe we can expose them if needed in the future.

sophokles73:
> > Ok with me! Only question is, do we really need to expose every symbol/const? I've really no idea, honest question.
> 
> At least for the UUri one, it's necessary for the transport layer (Zenoh, MQTT...). For others const, it depends. Maybe we can expose them if needed in the future.

That is actually a good question. @evshary Can you give an example where you need the constant itself, i.e. you do not need to do a check for which we already have a function on `UUri`?

sophokles73:
I guess what you would need is:
```rust
UUri::has_wildcard_entity_id(&self) -> bool;
UUri::has_wildcard_version(&self) -> bool;
UUri::has_wildcard_resource_id(&self) -> bool;
```
right?

evshary:
> I guess what you would need is:
> 
> ```rust
> UUri::has_wildcard_entity_id(&self) -> bool;
> UUri::has_wildcard_version(&self) -> bool;
> UUri::has_wildcard_resource_id(&self) -> bool;
> ```
> 
> right?

Correct. That's what I need :+1:

sophokles73:
I can quickly add these to UUri and then you can close this PR, ok?

evshary:
Close the PR since it's solved by https://github.com/eclipse-uprotocol/up-rust/pull/153

PR #152, opened 2024-07-16
Improve doc comments of L2 API types

sophokles73:


src/communication/default_notifier.rs
```diff
@@ -276,6 +284,7 @@ mod test {
         let notifier = SimpleNotifier::new(Arc::new(transport), uri_provider);
 
         let options = CallOptions::for_notification(None, None, None);
+        // resource ID of origin address must not be 0
```
AnotherDaniel, line 287: Is there maybe a spec line/requirement we could link this to?

AnotherDaniel (approved):
LGTM

src/communication/default_notifier.rs
sophokles73: I'd rather try to establish the link between the test case and the spec by means of an OpenFastTrace spec item (which will be added later) ...

Issue #154, opened 2024-07-16
Implement `Debug` for `ComparableListener` via its pointer to allow easier logging


PLeVasseur:
When I'm doing logging of data structures which hold a `ComparableListener` it'd be nice to have the ability to directly log that data structure using `{:?}` without having to destructure and do something special for `ComparableListener`.

PR #155, opened 2024-07-16
Add Debug impl for ComparableListener

PLeVasseur:
Addresses [#154]

### DIFF
--- a/src/utransport.rs
+++ b/src/utransport.rs
@@ -679,14 +679,9 @@ mod tests {
     fn test_comparable_listener_pointer_address() {
         let bar = Arc::new(ListenerBar);
         let comp_listener = ComparableListener::new(bar);
-        println!("comp_listener: {:?}", comp_listener);
 
         let comp_listener_thread = comp_listener.clone();
-        let handle = thread::spawn(move || {
-            println!("comp_listener_thread: {:?}", comp_listener_thread);
-
-            comp_listener_thread.pointer_address()
-        });
+        let handle = thread::spawn(move || comp_listener_thread.pointer_address());
 
         let comp_listener_address_other_thread = handle.join().unwrap();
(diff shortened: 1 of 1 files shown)


src/utransport.rs
```diff
@@ -656,4 +674,26 @@ mod tests {
         let check_on_receive_res = up_client_foo.check_on_receive(&uuri_1, &umessage);
         assert!(check_on_receive_res.is_err());
     }
+
+    #[test]
+    fn test_comparable_listener_pointer_address() {
+        let bar = Arc::new(ListenerBar);
+        let comp_listener = ComparableListener::new(bar);
+        println!("comp_listener: {:?}", comp_listener);
```
sophokles73, line 682: Do these println! statements contribute anything to the test?

PLeVasseur: They let me easily see the result if I run with --nocapture to "prove" the Debug implementation works :slightly_smiling_face: 

Would it be better to use logging crate (tracing?) and then env_logger to dev-dependencies?

PLeVasseur: I suppose I could do some string manipulation ensure we get the intended Debug implementation too, I suppose.

Which of those sounds reasonable?

sophokles73: Well, I have stated before that IMHO expecting the output of Display::display to contain any particular text is an antipattern.

That said, if you want to make sure that _something_ is produced, you could do something like
```rust
let bar = Arc::new(ListenerBar);
let comp_listener = ComparableListener::new(bar);
assert!(!format!("{}", comp_listener).is_empty());
```

PLeVasseur: Okay. I'll go with this.

sophokles73 (approved):
LGTM

Issue #15, opened 2023-12-21
Publish up-rust v0.1.0 to crates.io


PLeVasseur:
Hey @stevenhartley, @sophokles73, @dakrbo:

With the [issue](https://github.com/eclipse-uprotocol/uprotocol-rust/issues/10) closed regarding the implementation of the SDK, do you think now makes sense to publish a v0.1 to crates.io?

It's a pre 1.0 release, so we can still make breaking API changes like the one suggested to the [listener type definition](https://github.com/eclipse-uprotocol/uprotocol-rust/issues/14).

sophokles73:
Before we do that, we will need to perform an initial [Release Review](https://www.eclipse.org/projects/handbook/#release). After that, we can do subsequent releases for a year until the yearly *progress review*.

@stevenhartley you should be able to trigger the review using the [Eclipse Project Management Infrastructure (PMI)](https://projects.eclipse.org/projects/automotive.uprotocol/governance).

@dakrbo I would suggest to set the version of the crate to a SNAPSHOT version of the upcoming release in order to clearly point out that the code on the main branch is work in progress towards a release but does not reflect the release (yet). In our case that might be something like `0.1.5.SNAPSHOT`. When we do the release, i.e. create the corresponding tag in the repository, we set it to `0.1.5` and then change it to the next upcoming SNAPSHOT version, e.g. `0.1.6.SNAPSHOT`.

PLeVasseur:
Ah okay, good to learn about this. Thanks for the links!

Would it make sense to follow the SemVer spec? Gotta respect they [wrote down the BNF representation](https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions).

I suppose we'd be meeting SemVer spec if we did something like `0.1.5-SNAPSHOT`, using `SNAPSHOT` as our `<pre-release>` identifier.

What do you think?

PLeVasseur:
Maybe a separate question to @stevenhartley:

Does uProtocol version according to SemVer? In other words:
```
Bug fixes not affecting the API increment the patch version, backward compatible API additions/changes increment the minor version, and backward incompatible API changes increment the major version.
```

To where if e.g. the uProtocol spec changed to say, remove a uTransport interface or change the signature of one of them, would that entail a major bump? Feels kinda drastic while the details are still being worked out to me.

Perhaps the uProtocol spec could also follow a similar idea that @sophokles73 had regarding SNAPSHOT releases while things are being defined.

What do you think?

[...]

PR #156, opened 2024-07-17
Prepare for 0.1.0 release

sophokles73:
Several small fixes and improvements in preparation of the 0.1.0 release which is then to be published to crates.io.

This is for #15


src/utransport.rs
```diff
@@ -38,103 +38,45 @@ pub trait LocalUriProvider: Send + Sync {
...
-    /// block further message receipt. For long-running operations consider passing off received
-    /// data to a different async function to handle it and returning.
-    ///
-    /// # Note for `UTransport` implementers
-    ///
-    /// Because `on_receive()` is async you may choose to either `.await` it in the current context
-    /// or spawn it onto a new task and await there to allow current context to immediately continue.
+    /// This function is expected to return almost immediately. If it does not, it could potentially
+    /// block processing of succeeding messages. Long-running operations for processing a message should
+    /// therefore be run on a separate thread.
```
AnotherDaniel, line 58: '... thread' - or rather 'as a separate task'?

AnotherDaniel (approved):
LGTM, one minor editorial comment.

Excited!

src/utransport.rs
sophokles73: If you want to run blocking code, then you should probably use a dedicated (and separate) thread for running the code. Whether that code is scheduled as a task or run by means of message passing, is not relevant in this context FMPOV.

AnotherDaniel: The thought was that 'task' represents a bigger option space for that might be done - essentially the same argument that you're making, but my word associations between task and thread seem to be the other way around :-)

PR #157, opened 2024-07-17
Fix README file name in release workflow

sophokles73:


sophokles73:
@AnotherDaniel any other things we need to change to make this work?

AnotherDaniel (approved):
no brainer - which readme does cargo.toml reference?

sophokles73:
> no brainer - which readme does cargo.toml reference?

thanks for the pointer :+1: fixing that as well ...

PR #160, opened 2024-07-18
Use git submodule for proto3 files

sophokles73:
- added up-spec as git submodule.
- adapted build process to use proto3 files from submodule.
- added metadata for docs.rs build
- removed the "documentation" property from Cargo.toml in order
to automatically link to the generated docs on docs.rs.


evshary (approved):
LGTM :+1:

AnotherDaniel (approved):
Looking good! 
Not very related, but while we're cleaning up things: is this NOTICE.md file actually needed and relevant?

Issue #166, opened 2024-07-23
The combination of resource ID of source and sink


evshary:
When I'm using the L2 API, I found that `InMemoryRpcServer` call `register_listener` with source UUri `UUri::any()`, which means `//*/FFFF/FF/FFFF`.
https://github.com/eclipse-uprotocol/up-rust/blob/8a9fe59882c9ad67a63c7a3a052d37a5017082e2/src/communication/in_memory_rpc_server.rs#L232

I created a table [here](https://github.com/eclipse-uprotocol/up-spec/tree/main/up-l1#23-registerlistener) before to list all the combinations of resource ID for `register_listener`.
![image](https://github.com/user-attachments/assets/5dd318c8-773c-449e-b5c3-ad4a3187201a)
The reason I removed `source resource_id=FFFF & sink resource_id=[1-7FFF]` is that the source resource_id should always be 0 here. However, I agree that in the L2 API case, it might be annoying for users to use `//*/FFFF/FF/0` as the source Uri.

Furthermore, if we decided to add back the combination of `{FFFF, 1-7FFF}`, then it will come across [the question I met before](https://github.com/eclipse-uprotocol/up-spec/pull/188#discussion_r1673355607). Should we also add back the other cases?

@sophokles73 WDYT?

sophokles73:
I agree that in line 232 we should use `//*/FFFF/FF/0` as the origin filter if not specified explicitly by the user.

sophokles73:
Would you like to create a PR?

PR #167, opened 2024-07-24
Set resource ID to 0 while listening in InMemoryRpcServer.

evshary:
The PR is to deal with https://github.com/eclipse-uprotocol/up-rust/issues/166.
If we don't configure the origin_filter, we'll use "//*/FFFF/FF/0" as the default.

### DIFF
--- a/src/communication/in_memory_rpc_server.rs
+++ b/src/communication/in_memory_rpc_server.rs
@@ -229,7 +229,9 @@ impl RpcServer for InMemoryRpcServer {
             });
             self.transport
                 .register_listener(
-                    origin_filter.unwrap_or(&UUri::any_with_resource_id(0)),
+                    origin_filter.unwrap_or(&UUri::any_with_resource_id(
+                        crate::uri::RESOURCE_ID_RESPONSE,
+                    )),
                     Some(&sink_filter),
                     listener.clone(),
                 )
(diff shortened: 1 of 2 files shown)


src/uri.rs
```diff
@@ -380,6 +380,17 @@ impl UUri {
         }
     }
 
+    /// Gets a URI that consists of wildcards and the specific resource ID.
+    pub fn any_with_resource_id(resource_id: u32) -> Self {
```
sophokles73, line 384: how about changing the implementation of `Self::any` to
```rust
Self::any_with_resource_id(WILDCARD_RESOURCE_ID)
```
in order to reduce code duplication?

src/communication/in_memory_rpc_server.rs
```diff
@@ -229,7 +229,7 @@ impl RpcServer for InMemoryRpcServer {
             });
             self.transport
                 .register_listener(
-                    origin_filter.unwrap_or(&UUri::any()),
+                    origin_filter.unwrap_or(&UUri::any_with_resource_id(0)),
```
sophokles73, line 232: maybe you can use the const?
```
origin_filter.unwrap_or(&UUri::any_with_resource_id(crate::uri::RESOURCE_ID_RESPONSE)),
```

src/uri.rs
evshary: Good idea.

src/communication/in_memory_rpc_server.rs
evshary: Sure

sophokles73 (approved):
LGTM

PR #169, opened 2024-07-25
Empty Authority field implies local UUri

AnotherDaniel:
UUris with empty Authority are considered to be local - this was missing in the is_remote_uuri check.

### DIFF
--- a/src/uri.rs
+++ b/src/uri.rs
@@ -389,7 +389,8 @@ impl UUri {
         self.eq(&UUri::default())
     }
 
-    /// Check if an `UUri` is remote, by comparing authority fields.
+    /// Check if an `UUri` is remote, by comparing authority fields. UUris with empty authority are
+    /// considered to be local.
     ///
     /// # Returns
     ///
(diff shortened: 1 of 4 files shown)


src/uri.rs
```diff
@@ -406,7 +407,8 @@ impl UUri {
     /// assert!(authority_a.is_remote_authority(&authority_b));
     /// ````
     pub fn is_remote_authority(&self, other_uri: &UUri) -> bool {
-        self.authority_name != WILDCARD_AUTHORITY
+        !self.authority_name.is_empty()
```
sophokles73, line 410: a unit test would be nice, also covering the other cases ...

AnotherDaniel: Done

src/uri.rs
```diff
@@ -404,9 +405,18 @@ impl UUri {
     /// let authority_a = UUri::from_str("up://Authority.A/100A/1/0").unwrap();
     /// let authority_b = UUri::from_str("up://Authority.B/200B/2/20").unwrap();
     /// assert!(authority_a.is_remote_authority(&authority_b));
+    ///
+    /// let authority_local = UUri::from_str("up:///100A/1/0").unwrap();
+    /// assert!(!authority_local.is_remote_authority(&authority_a));
+    ///
+    /// let authority_wildcard = UUri::from_str("up://*/100A/1/0").unwrap();
+    /// assert!(!authority_local.is_remote_authority(&authority_a));
```
sophokles73, line 413: this seems to be a duplicate of 410

AnotherDaniel: Thx for spotting this - copy&pasty...

sophokles73 (approved):
LGTM

Issue #178, opened 2024-08-07
Add SubscriptionInformation to up-rust usubscription


matthewd0123:
Hi all!

I have been working on a PR for up-streamer-rust to add the ability for the streamer to track subscriptions, so it can properly route publish-type messages. As part of this, I have created a struct that is used to store and track this information. @AnotherDaniel thought it might be good to keep this information in up-rust instead of directly in the streamer code. I will open a PR with this information. Once the streamer PR is merged, would ask that this code get reviewed.

PR #179, opened 2024-08-07
Add SubscriptionInformation struct to usubscription

matthewd0123:
* In uStreamer, there will be a SubscriptionCache that is used to locally track subscriptions coming in from the subscription service. We needed to create a data structure to track these subscriptions, which I have defined here.

### DIFF
--- a/src/core/usubscription.rs
+++ b/src/core/usubscription.rs
@@ -26,6 +26,41 @@ pub use crate::up_core_api::usubscription::{
 
 use crate::{UStatus, UUri};
 
+// Tracks information for the SubscriptionCache
+pub struct SubscriptionInformation {
+    pub topic: UUri,
+    pub subscriber: SubscriberInfo,
+    pub status: SubscriptionStatus,
+    pub attributes: SubscribeAttributes,
+    pub config: EventDeliveryConfig,
+}
+
+impl Eq for SubscriptionInformation {}
+
+impl PartialEq for SubscriptionInformation {
+    fn eq(&self, other: &Self) -> bool {
+        self.subscriber == other.subscriber
+    }
+}
+
+impl Hash for SubscriptionInformation {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.subscriber.hash(state);
+    }
+}
+
+impl Clone for SubscriptionInformation {
+    fn clone(&self) -> Self {
+        Self {
+            topic: self.topic.clone(),
+            subscriber: self.subscriber.clone(),
+            status: self.status.clone(),
+            attributes: self.attributes.clone(),
+            config: self.config.clone(),
+        }
+    }
+}
+
 impl Hash for SubscriberInfo {
     /// Creates a hash value based on the URI property.
     ///


matthewd0123:
#178

src/core/usubscription.rs
```diff
@@ -26,6 +26,41 @@ pub use crate::up_core_api::usubscription::{
 
 use crate::{UStatus, UUri};
 
+// Tracks information for the SubscriptionCache
```
PLeVasseur, line 29: Use triple forward slashes for doc comments
```suggestion
/// Tracks information for the SubscriptionCache
```

src/core/usubscription.rs
```diff
@@ -26,6 +26,41 @@ pub use crate::up_core_api::usubscription::{
...
+
+impl Eq for SubscriptionInformation {}
+
+impl PartialEq for SubscriptionInformation {
+    fn eq(&self, other: &Self) -> bool {
+        self.subscriber == other.subscriber
+    }
+}
+
+impl Hash for SubscriptionInformation {
```
PLeVasseur, line 46: Suggest adding similar level of doc comment and doc example as [below](https://github.com/eclipse-uprotocol/up-rust/pull/179/files#diff-baef08640da88466a243494249b77b5686bb03c4f5a77017a6a327ff0aa18678R94)

src/core/usubscription.rs
```diff
@@ -26,6 +26,41 @@ pub use crate::up_core_api::usubscription::{
...
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.subscriber.hash(state);
+    }
+}
+
+impl Clone for SubscriptionInformation {
```
PLeVasseur, line 52: Do we need to explicitly `impl Clone` in this case, since all the members are just having `.clone()` in turn called on them?

Reference [this Rust Playground snippet](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5b7d466181097578531614df33bb3c2d).

Would it suffice to derive `Clone` for `SubscriptionInformation` as shown in the snippet?

src/core/usubscription.rs
```diff
@@ -26,6 +26,41 @@ pub use crate::up_core_api::usubscription::{
...
+    pub topic: UUri,
+    pub subscriber: SubscriberInfo,
+    pub status: SubscriptionStatus,
+    pub attributes: SubscribeAttributes,
+    pub config: EventDeliveryConfig,
+}
+
+impl Eq for SubscriptionInformation {}
+
+impl PartialEq for SubscriptionInformation {
```
PLeVasseur, line 40: I think some comment explaining the choice to use `subscriber` as the only member to check equality on would be good.

src/core/usubscription.rs
PLeVasseur, line 46: Same sort of comment as above -- good to explain here why only the `subscriber` is being used in the Hash.

src/core/usubscription.rs
PLeVasseur: Would be good to expand this a bit with a code usage example a in triple tick block, e.g.
```rust
/// ```
/// // put the usage here
/// ```
```

that way it'll be compiled and tested.

PLeVasseur (changes requested):
Take a look at some suggestions. Overall would like to see more doc comments / doc tests

sophokles73:
Is this struct helpful to other users than uStreamer? If not, I do not see why it should go into up-rust...

@PLeVasseur WDYT?

PLeVasseur:
Let's close for now. There needs to be a renewed effort in how to properly do subscription interactions within the uStreamer.

Issue #180, opened 2024-08-12
Add nightly check for compatibility with up-spec's main branch


sophokles73:
We should add a workflow that tries to compile the up-rust crate against the `HEAD` revision of the `up-spec` repository (usually that would be the top of the `main` branch).
This will allow us to detect any breaking changes introduced in the spec early on.

sophokles73:
closed via #184

PR #181, opened 2024-08-12
Add nightly check for compatibility with up-spec

sophokles73:
addresses #180


evshary (approved):
LGTM
Just out of curiosity, do you think we should put it inside the `nightly.yaml`?
Or maybe mention the workflow in the [README](https://github.com/eclipse-uprotocol/up-rust/blob/main/.github/workflows/README.md)?

sophokles73:
> LGTM Just out of curiosity, do you think we should put it inside the `nightly.yaml`? 

Since I expect this to fail (often?) I'd rather not have it taint the results of the Nightly build which is used to make sure that our main branch builds correctly.

> Or maybe mention the workflow in the [README](https://github.com/eclipse-uprotocol/up-rust/blob/main/.github/workflows/README.md)?

Yes, that's a good idea. I will add it there ...

Issue #182, opened 2024-08-12
Add workflow for tracing requirements from up-spec


sophokles73:
Many of the artifacts in the up-spec repository already contain OpenFastTrace specification items which can be traced to implementations and tests using the [newly added AsciiDoc Plugin](https://github.com/itsallcode/openfasttrace-asciidoc-plugin) for OpenFastTrace.

We should be able to add a GitHub workflow that runs the tracer and allows the resulting report to be added as a release artifact.

sophokles73:
closed via #183

PR #183, opened 2024-08-12
Add workflow for running OpenFastTrace against up-spec

sophokles73:
Addresses #182


sophokles73:
@evshary can you take a look? I think I got it now ...

evshary (approved):
LGTM

PR #184, opened 2024-08-12
Include requirements tracing in nightly up-spec compatibility check

sophokles73:
Refactored OFT execution into its own composite GitHub Action in order
to be able to use it with both, the latest released up-spec as well as
the latest up-spec from the main branch.

This is for #180


evshary (approved):
LGTM

PR #188, opened 2024-08-13
Exclude L1 and L3 requirements from OFT check

sophokles73:
The L1 and L3 specifications contain requirements for implementations
of the corresponding APIs and thus are not relevant for the Rust
language library, which only defines these APIs but does not
implement them.

Note that the library contains a (default) implementation of the L2
API and thus still needs to cover the requirements specified for L2.


PLeVasseur (approved):
LGTM

PR #190, opened 2024-08-15
Use GitHub variable to define OFT file patterns

sophokles73:
Use UP_SPEC_OPEN_FAST_TRACE_FILE_PATTERNS variable to look up the file patterns to use for running OFT.

Also adapted the run-oft Action to dynamically determine, if a JRE needs to be installed.


evshary (approved):
LGTM

PR #191, opened 2024-08-15
Install JDK if necessary only

sophokles73:
The run-oft Action now dynamically determines if the JRE that is
installed already is recent enough to run OFT.
Only if that is not the case, JDK 22 is being installed.


evshary (approved):
LGTM

PR #193, opened 2024-08-21
Use standard OpenFastTrace GitHub Action

sophokles73:
Replace local implementation of OFT action with standard Action provided by the OpenFastTrace project.


evshary (approved):
LGTM

PR #194, opened 2024-08-21
Fix OFT report file format

sophokles73:
The local composite action has been adapted to explicitly  set the
report format to "html" when invoking the standard OFT Action to
override the default ("plain").


evshary (approved):
LGTM

PR #195, opened 2024-08-21
Improve usability of UPayload

sophokles73:
Adapted constructor to accept any payload that is Into<Bytes>.

### DIFF
--- a/src/communication.rs
+++ b/src/communication.rs
@@ -254,10 +254,22 @@ pub struct UPayload {
 
 impl UPayload {
     /// Creates a new payload for some data.
-    pub fn new(payload: Bytes, payload_format: UPayloadFormat) -> Self {
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use up_rust::UPayloadFormat;
+    /// use up_rust::communication::UPayload;
+    ///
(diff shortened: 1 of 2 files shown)


src/communication.rs
```diff
@@ -254,10 +254,22 @@ pub struct UPayload {
...
+    /// ```rust
+    /// use up_rust::UPayloadFormat;
+    /// use up_rust::communication::UPayload;
+    ///
+    /// let data: Vec<u8> = vec![0x00_u8, 0x01_u8, 0x02_u8];
+    /// let payload = UPayload::new(data, UPayloadFormat::UPAYLOAD_FORMAT_RAW);
+    /// assert_eq!(payload.payload_format(), UPayloadFormat::UPAYLOAD_FORMAT_RAW);
+    /// assert_eq!(payload.payload().len(), 3);
+    /// ```
+    pub fn new<T: Into<Bytes>>(payload: T, payload_format: UPayloadFormat) -> Self {
```
PLeVasseur, line 269: Nice, allows us to accept `Bytes` as well as anything which can be turned into `Bytes`.

PLeVasseur (approved):
LGTM

PR #197, opened 2024-08-23
Use separate repo variable for up-rust file patterns

sophokles73:
The workflows that run OpenFastTrace for checking coverage of
requirements from up-spec have been adapted to use a separate
repo variable that contains the file patterns for artifacts from
up-rust that cover the requirements.


evshary (approved):
LGTM

PR #199, opened 2024-08-23
Prepare 0.2.0 release

sophokles73:
Added some new behavior so using a new minor version seems
reasonable, in order to make this more transparent.


AnotherDaniel (approved):
LGTM

PR #201, opened 2024-09-12
Update cargo deps to address box_syntax lint

AnotherDaniel:
There is a new lint concerning the discontinuation of `#![allow(box_pointers)]`, which is used by protoc-generated code.
(also refer to https://github.com/rust-lang/rust/issues/49733)

This updates our deps , including the relevant proto generation libraries, that gets rid of this issue.


PLeVasseur (approved):
LGTM

PR #202, opened 2024-09-12
Spec sync

AnotherDaniel:
Align with recent changes in usubscription object definition in up-spec; these break the uP Spec Compatibility CI

### DIFF
--- a/src/communication/default_pubsub.rs
+++ b/src/communication/default_pubsub.rs
@@ -24,7 +24,7 @@ use tracing::{debug, info};
 
 use crate::{
     core::usubscription::{
-        self, State, SubscriberInfo, SubscriptionRequest, USubscription, UnsubscribeRequest, Update,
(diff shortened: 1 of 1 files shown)


sophokles73:
If we merge this, then we can no longer create releases based on up-spec 1.6.0-alpha.3, because some of the changes introduced to up-spec since that tag are not backwards compatible.
If we want to start adopting those changes, then I would propose to do so on a separate branch, which we could then (repeatedly) rebase on HEAD of (up-rust) `main`, if necessary.
WDYT @AnotherDaniel ?

sophokles73:
FMPOV the first (separate) commit on that branch should be the forwarding of the up-spec submodule ...

AnotherDaniel:
Will deal with this by publishing a dedicated branch (1.6.0-alpha.4_changes)

Issue #203, opened 2024-09-24
Add feature for mapping between UMessage and CloudEvent


sophokles73:
The [uProtocol Spec](https://github.com/eclipse-uprotocol/up-spec/blob/main/up-l1/cloudevents.adoc) defines a canonical mapping between uMessages and CloudEvents.
up-rust should support the mapping using the Protobuf Format for CloudEvents by means of an (optional) feature.

sophokles73:
closed via #216 and #217

PR #204, opened 2024-09-26
Support mapping between UMessage and cloudevents

sophokles73:
An optional feature has been added which provides functions for mapping
UMessages to corresponding CloudEvents and vice versa as defined in
the uProtocol specification.

Addresses #203]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,6 +27,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "android-tzdata"
+version = "0.1.1"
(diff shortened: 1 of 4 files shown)


sophokles73:
Hmm, looks like there is some effort going on to update many of the cloudevents-sdk crate's dependencies to more recent versions: https://github.com/cloudevents/sdk-rust/pull/233

I guess it makes sense to wait for that to happen ...

sophokles73:
Feature has already been added to up-rust

PR #205, opened 2024-10-02
Add simple LocalUriProvider implementation

sophokles73:
Added StaticUriProvider which is configured using static information.
StaticUriProvider is supposed to be used by transport implementations
and/or test cases.


evshary (approved):
Looks cleaner to me

PR #206, opened 2024-10-02
Use StaticUriProvider in examples

sophokles73:
The examples have been changed to use StaticUriProvider instead of
having LocalTransport implement LocalUriProvider.


evshary (approved):
LGTM

Issue #207, opened 2024-10-10
Nightly build fails due to MSRV check


sophokles73:
The nightly job fails because the MSRV check fails to execute.

sophokles73:
closed via #208

PR #208, opened 2024-10-10
Fix MSRV check in nightly build

sophokles73:
The "--verify" parameter has been removed in version 0.16 of the
msrv-check cargo plugin. The command line to run the check has been
adapted accordingly.

addresses #207


AnotherDaniel (approved):
LGTM

Issue #209, opened 2024-11-07
Did we remove the validation for ue_version_major and resource_id being the correct number of bits?


PLeVasseur:
I was looking to use this today and seems like it was not migrated over when we had core API changes a while back.

I see in [here](https://github.com/eclipse-uprotocol/up-rust/commit/57c3c8c1b69880e469529424431cc3805b006576#diff-c268be64cebcd98764bc192af60854a8ae20ec005a47cd379314f36b6f0cc57a) @sophokles73 seems to have removed (?) the concept of validating the length of bits contained in the `ue_version_major` and `resource_id`.

Maybe I'm missing something though?

I like how [`UUri::try_from_parts`](https://github.com/eclipse-uprotocol/up-rust/blob/614e73aae8b78bdd14c09449dc8bbe898c5ace6d/src/uri.rs#L369) will give us a correct-by-construction `UUri`, but I'd like the ability to verify that a `UUri` that I'm given matches the correct specification [here](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc#2-data-model).

Perhaps we removed the ability to check this?

Can I get a sanity check here before I reimplement this?

sophokles73:
Hmm, when parsing a UUri from a URI string we do check that it is an 8 bit unsigned integer. But when we serialize to a URI string we do not. I guess you are right in that we should check there too. Is that what you are concerned about?

PLeVasseur:
Yeah, validating a `UUri` that's of unknown provenance is what I was getting at. I think you've understood my ask and are working to implement the ability to do that in #210.

PR #210, opened 2024-11-12
Add validty check function to UUri

sophokles73:
Added a function for checking if an existing UUri is a valid
uProtocol URI.

Addresses #209

### DIFF
--- a/src/uri.rs
+++ b/src/uri.rs
@@ -161,26 +161,7 @@ impl FromStr for UUri {
         }
         let authority_name = parsed_uri
             .authority()
-            .map_or(Ok(String::default()), |auth| {
-                if auth.has_port() {
-                    Err(UUriError::serialization_error(
-                        "uProtocol URI's authority must not contain port",
-                    ))
-                } else if auth.has_username() || auth.has_password() {
-                    Err(UUriError::serialization_error(
-                        "uProtocol URI's authority must not contain userinfo",
-                    ))
-                } else {
-                    let auth_name = auth.host().to_string();
-                    if auth_name.len() <= 128 {
-                        Ok(auth_name)
-                    } else {
-                        Err(UUriError::serialization_error(
-                            "URI's authority name must not exceed 128 characters",
-                        ))
-                    }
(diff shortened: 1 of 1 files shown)


sophokles73:
@PLeVasseur is this what you need for uStreamer?

src/uri.rs
```diff
@@ -419,6 +382,77 @@ impl UUri {
...
+                Ok(auth_name)
+            } else {
+                Err(UUriError::validation_error(
+                    "URI's authority name must not exceed 128 characters",
+                ))
+            }
+        }
+    }
+
+    fn verify_major_version(major_version: u32) -> Result<u8, UUriError> {
```
PLeVasseur, line 416: :+1: Yup, this is definitely one of the functions that'd be useful to have.

src/uri.rs
```diff
@@ -419,6 +382,77 @@ impl UUri {
...
+    ///   ue_id: 0x1000,
+    ///   ue_version_major: 0x01,
+    ///   resource_id: 0x8100,
+    ///   ..Default::default()
+    /// };
+    /// assert!(uuri.check_validity().is_ok());
+    /// ```
+    pub fn check_validity(&self) -> Result<(), UUriError> {
+        Self::verify_authority(self.authority_name.as_str())?;
+        Self::verify_major_version(self.ue_version_major)?;
```
PLeVasseur, line 452: Could you also add a `verify_resource_id()` that'd attempt the conversion to a `u16` and include calling that function in here too?
image

PLeVasseur (commented):
Thanks  @sophokles73!

I think this is one half of what I was asking for, please see comments.

src/uri.rs
sophokles73: Oh, absolutely. Don't know how that slipped through. Good that we do the reviews ... ;-)

sophokles73:
@PLeVasseur I have added a corresponding check ...

PLeVasseur (approved):
Thanks Kai! LGTM

Issue #186, opened 2024-08-13
Adopt changes to UUri instance ID wildcard


sophokles73:
The uProtocol UUri specification has been [changed](https://github.com/eclipse-uprotocol/up-spec/pull/222) to use `0xFFFF` as the wildcard for a service's instance ID instead of `0x0000`.

up-rust needs to be adapted accordingly once up-spec `1.6.0-alpha.4` has been released.

sophokles73:
closed via #211

PR #211, opened 2024-11-12
Adopt changes to UUri wildcards

sophokles73:
UUri has been changed to use FFFF as the pattern for matching any
entity instance.

Addresses #186


AnotherDaniel (approved):
LGTM!

Issue #185, opened 2024-08-12
Expose MockTransport for use in other repositories


matthewd0123:
Hello,

I've been working on up-streamer-rust, and in the course of writing some tests, I found the need to mock some uTransports to test error conditions. I saw MockTransport in https://github.com/eclipse-uprotocol/up-rust/blob/main/src/utransport.rs and think it would be very useful to have in that repo, so am proposing to make that a pub. If that's acceptable, I will make the PR for it.

sophokles73:
A PR would be very welcome :-)
I would suggest to make it conditional on either the `test` or `util` profile then ...

sophokles73:
closed via #213

PR #213, opened 2024-11-15
Add feature for exposing mock implementations

sophokles73:
The "test-util" feature has been added which exposed the mockall based
mock implementations of the transport and communication level API
traits. These can be helpful to e.g. crates implementing UTransport
for implementing unit tests.

Addresses #185


PLeVasseur (approved):
LGTM

All-in-all seems reasonable :)

PR #220, opened 2024-12-04
Add getters for URI components

### LINKED ISSUE #219: Provide a way to get entity type from UUri
Just as we discussed here https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/pull/93#discussion_r1867312701, it would be convenient if we could provide a function for transport layers to get the entity type from UUri.

sophokles73:
Added convenience functions for accessing a UUri's sub-component values
using the data types as defined in the uProtocol specification.

Fixes #219

### DIFF
--- a/src/uri.rs
+++ b/src/uri.rs
@@ -384,6 +384,76 @@ impl UUri {
         }
     }
 
+    /// Gets the authority name part from this uProtocol URI.
+    ///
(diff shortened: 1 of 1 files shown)


evshary (approved):
LGTM
It's good to have other getters for all URI components

PR #227, opened 2025-01-14
Improve MSRV checking

sophokles73:
Updated GitHub workflow for checking the MSRV to use the cargo-hack tool
which is also recommended by the Cargo book.
Also updated crate's MSRV to 1.76 because the "tests" target failed the
check for 1.74.1.

The nightly workflow has been extended with a job that verifies if the
crate can be built with the latest (compatible) versions of all
dependencies. The job will not fail the overall workflow execution
but provides an indication that the crate's MSRV might need to be
adapted.


AnotherDaniel (approved):
LGTM

Issue #230, opened 2025-02-11
Nightly build fails for checking build with latest deps


sophokles73:
```
Running `/home/runner/work/up-rust/up-rust/target/debug/build/up-rust-de0e715e8fee93af/build-script-build`
error: failed to run custom build command for `up-rust v0.4.0 (/home/runner/work/up-rust/up-rust)`

Caused by:
  process didn't exit successfully: `/home/runner/work/up-rust/up-rust/target/debug/build/up-rust-de0e715e8fee93af/build-script-build` (exit status: 1)
  --- stderr
  up-spec/up-core-api/: warning: directory does not exist.
  Could not make proto path relative: up-spec/up-core-api/uprotocol/uoptions.proto: No such file or directory
  codegen failed: parse and typecheck

  Caused by:
      0: using protoc parser
      1: protoc command "/home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/protoc-bin-vendored-linux-x86_64-3.1.0/bin/protoc" "-Iup-spec/up-core-api/" "--descriptor_set_out=/tmp/protobuf-parsee21G23/descriptor.pbbin" "--include_imports" "up-spec/up-core-api/uprotocol/uoptions.proto" "up-spec/up-core-api/uprotocol/v1/uuid.proto" "up-spec/up-core-api/uprotocol/v1/uri.proto" "up-spec/up-core-api/uprotocol/v1/uattributes.proto" "up-spec/up-core-api/uprotocol/v1/ucode.proto" "up-spec/up-core-api/uprotocol/v1/umessage.proto" "up-spec/up-core-api/uprotocol/v1/ustatus.proto" "up-spec/up-core-api/uprotocol/v1/file.proto" "up-spec/up-core-api/uprotocol/core/udiscovery/v3/udiscovery.proto" "up-spec/up-core-api/uprotocol/core/usubscription/v3/usubscription.proto" "up-spec/up-core-api/uprotocol/core/utwin/v2/utwin.proto" exited with non-zero code
```

sophokles73:
closed via #231

PR #231, opened 2025-02-11
Recurse submodules when checking out repo

sophokles73:
The workflow for checking if the code can be compiled using the latest versions of 3rd party dependencies had failed sporadically because the uProtocol API proto3 files had not been available.

This was due to the fact that the up-spec submodule had not been properly initialized during checkout.

Addresses #230


AnotherDaniel (approved):
LGTM

Issue #232, opened 2025-02-17
Use assets from CI/CD repository


sophokles73:
With the advent of the [uProtocol CI/CD repository](https://github.com/eclipse-uprotocol/ci-cd), many of the workflows included in up-rust should be replaced by references to the assets in the CI/CD repo.

sophokles73:
closed via #233, #234 and #235

PR #233, opened 2025-02-17
Replace local workflows

sophokles73:
Use references to assets from the uProtocol CI/CD repo instead of
keeping local copies of workflows/actions.

Addresses #232


sophokles73:
@AnotherDaniel would you mind taking a look?

PR #234, opened 2025-02-17
Improve CI job for running tests

sophokles73:
Adapted the execution of cargo nextest to use a CI specific profile and
produce test results in JUnit format.
Included the running of the doctests in the same job that runs cargo
nextest so that the doctests are also executed for all relevant
feature combinations.
Merged the test results into a single archive so that all results are
part of the archive that is attached to releases.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -31,9 +31,6 @@ on:
       test_results_url:
         description: "URL of the test results artifact"
         value: ${{ jobs.nextest.outputs.test_results_url }}
-      doctest_results_url:
-        description: "URL of the doctest results artifact"
-        value: ${{ jobs.doctest.outputs.test_results_url }}
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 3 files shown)


.github/workflows/check.yaml
```diff
@@ -155,41 +152,27 @@ jobs:
...
 
       - name: Upload all-features test results artifact
         id: test_results
         if: matrix.feature-flags == '--all-features'
         uses: actions/upload-artifact@v4
         with:
           name: test-results
-          path: testresults--all-features.json
-
-  # [impl->req~up-language-ci-test~1]
```
AnotherDaniel, line 172: In this scenario, where did you move the req implementation marker?

AnotherDaniel (commented):
Minor question

.github/workflows/check.yaml
sophokles73: It is already on the `nextest` job which now also includes running the doctests ...

AnotherDaniel: Ah I see

AnotherDaniel (approved):
LGTM

PR #236, opened 2025-02-21
Fix typos

sophokles73:


AnotherDaniel (approved):
I feel strong sympathy about these type patterns

PR #237, opened 2025-02-21
Improve CI/CD worklfows

sophokles73:


.github/workflows/nightly.yaml
```diff
@@ -47,6 +47,16 @@ jobs:
         command: check
         arguments: --all-features
 
+  audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: "recursive"
+      - uses: rustsec/audit-check@v2
```
AnotherDaniel, line 56: Is cargo-audit really adding any benefit for us, beyond what cargo-deny does (https://github.com/EmbarkStudios/cargo-deny/issues/386#issuecomment-995151508)

sophokles73: Good point, I didn't realize that cargo-deny already has advisory checking and is actually much easier to configure :-)

sophokles73:
Closed in favor of relying on `cargo-deny`

PR #239, opened 2025-02-25
Improve cargo deny check run during nightly

sophokles73:
Use locked dependencies when performing the check in order to get
reproducible results. Also exclude dev dependencies from checks.


AnotherDaniel (approved):
LGTM

PR #241, opened 2025-03-11
Move expiration check to UAttributes

sophokles73:
The function for checking if a message has expired already has been
moved from UAttributesValidator to UAttributes in order to increase
cohesion and encapsulation.


AnotherDaniel (approved):
LGTM

PR #242, opened 2025-03-11
Add convenience functions for reading UMessage fields

sophokles73:
Added accessor functions for reading fields of UMessage without the
need to acccess the attributes, thus increasing encapsulation and
providing a much more convenient API for client code.


AnotherDaniel (approved):
Much nicer

PLeVasseur (approved):
Mmm, this is nice.

PR #243, opened 2025-03-12
Prepare 0.5.0 release

sophokles73:


AnotherDaniel (approved):
Looking good; one question: there are some version bumps that code be made - for bytes, rand, thiserror, tokio - should we take this opportunity?

sophokles73:
Didn't we have a discussion about the overall strategy for the MSRV we want to follow?
IIRC we decided to generally keep close to the _latest & greatest_ and to not pay too much attention on backwards compatibility (yet). Transposing this to the dependencies, what you say makes sense to me :-)
I think we can safely use more recent minor versions whenever we already use a major version > 0, assuming that the components actually follow semver practices ...

PR #244, opened 2025-03-12
Increase MSRV

sophokles73:
to enable more recent language constructs like Option::is_none_or
which make the code easier to write (and understand).

This should be in line with our general practice of keeping close
to the most recent Rust version.


AnotherDaniel (approved):
LGTM

PR #246, opened 2025-03-21
Use OFT Tags to filter relevant specification items

sophokles73:
Also adopted common GitHub VAR name specifying artifacts to include
in the OFT run.


AnotherDaniel (approved):
approved

PR #247, opened 2025-03-21
Use OFT run's exit code as job's exit code

sophokles73:
Made sure that the job for checking up-spec compliance fails if either
the tests fail or any of the relevant requirements from up-spec are
not covered.
This way we get an "early warning" if the current up-rust code base
does not comply (yet) with upcoming changes in the spec.


AnotherDaniel (approved):
LGTM

PR #248, opened 2025-03-27
Fix nightly execution of requirements tracing

sophokles73:
The check has failed recently because it had been erroneously using
the file patterns that refer to the current (not yet released) state
of up-spec.

Also replaced GitHub repo variables with explicit specification of
file patterns in order to have more flexibility when defining the
scope of the check.

### DIFF
--- a/.github/workflows/check-up-spec-compatibility.yaml
+++ b/.github/workflows/check-up-spec-compatibility.yaml
@@ -54,7 +54,7 @@ jobs:
         id: run-oft
         uses: eclipse-uprotocol/ci-cd/.github/actions/run-oft@main
         with:
-          file-patterns: "${{ vars.UP_SPEC_OPEN_FAST_TRACE_FILE_PATTERNS }} ${{ vars.UP_COMPONENT_OPEN_FAST_TRACE_FILE_PATTERNS }}"
+          file-patterns: "up-spec/*.adoc up-spec/*.md up-spec/basics up-spec/up-l1/README.adoc up-spec/up-l1/cloudevents.adoc up-spec/up-l2/api.adoc *.adoc *.md *.rs .github examples src tests tools"
           tags: "_,LanguageLibrary"
 
       # now try to build and run the tests which may fail if incomaptible changes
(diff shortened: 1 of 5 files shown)


AnotherDaniel (approved):
Looks good

sophokles73:
@AnotherDaniel would you mind taking another look? I have added some changes making the distinction between _compliance with current spec_ and _compatibility with latest spec_ more explicit.

AnotherDaniel (commented):
I like the differentiation you're introducing here, between the Programmatic and the Conceptual.

One question: is there a better place to define the file patterns? I really have not thought this through, first thing that comes to mind: is this information actually worth putting in an explicit little file in the repo, so that it be defined and managed together with the code it pertains to?  (I realise that this is kind of the case now, inline in the action definition. But I think this is somewhat obscure, plus changes to the file-pattern scope shouldn't necessarily result in changes to the CI pipeline itself...)

sophokles73:
> One question: is there a better place to define the file patterns? I really have not thought this through, first thing that comes to mind: is this information actually worth putting in an explicit little file in the repo, so that it be defined and managed together with the code it pertains to?

That is exactly what I tried to achieve by having the respective file patterns being defined in one place only: the two specific workflows. The _compliance_ check uses (potentially) different files than the _compatibility_ check (which needs to consider the latest developments in up-spec).

AnotherDaniel:
I amended my comment while you wrote yours :)

The little nag I feel is that, it feels not optimal to have to modify CI pipeline files (infrastructure) for this kind of scan-scope change. 
But that might just be me...

sophokles73:
> I amended my comment while you wrote yours :)
> 
> The little nag I feel is that, it feels not optimal to have to modify CI pipeline files (infrastructure) for this kind of scan-scope change. But that might just be me...

So you would prefer to read the patterns from text files in the repo?

AnotherDaniel:
I'd at least like to get your thoughts on separating this out from the CI scripts. I'm not certain how strongly I feel about this right now.

AnotherDaniel:
Ok fair enough.

PR #252, opened 2025-04-08
Use Gherkin to define TCK for UUID and UUri serialization

sophokles73:
The up-tck component had been using Gherkin for defining the TCK's
test scenarios. The idea was that these could be used for verifying a
language library's compliance with the specification.

This is a first step in the direction described in
https://github.com/eclipse-uprotocol/up-tck/issues/116.

The TCK's test scenarios have been adapted to cover more cases and, in
particular, also cover failure cases. The corresponding unit tests have
been replaced by Cucumber based tests and the Gherkin files have been
amended with OpenFastTrace specification identifiers and are now also
included when doing the requirements tracing run.

This is mainly a demonstration of how this could work. If we decide to go in this direction, we would first need to move the Gherkin files to up-spec, of course.


features/uuri/pattern_matching.feature
```diff
@@ -0,0 +1,57 @@
...
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-FileType: SOURCE
+# SPDX-License-Identifier: Apache-2.0
+#
+Feature: Matching endpoint identifiers (UUri) against patterns
+
+  Scenario Outline:
+    Developers using a uProtocol language library should be able to verify that a specific
+    endpoint identifier matches a given pattern as specified by the UUri specification.
```
AnotherDaniel, line 18: Is it possible to add links (to the spec) here? Probably something to do once we move this over to up-spec...

AnotherDaniel (approved):
I like this - looked into cucumber a year ago when beginning to rewrite up-subscription spec, but as usual you beat me to it for actually using the thing.

features/uuri/pattern_matching.feature
sophokles73: Instead of linking directly to the corresponding sections in the spec documents, I would rather rely on the OFT specification items which are used to link to the corresponding requirements defined in the specification documents. Links tend to become invalid over time without anybody noticing. However, if the OFT spec items are changed, we will immediately know when doing the trace run ...

sophokles73:
I also had to adapt the _check.yaml_ GitHub workflow to correctly run all tests (unit/doc/integration) and create junit reports for them.

I have also figured out how to include the integration (TCK) tests in the tarpaulin (test coverage) run. That will require a (small) change to the shared workflow definition in the uprotocol/ci-cd repo. We can then replace the checks that we had originally implemented in unit tests with the corresponding TCK tests ...

sophokles73:
@PLeVasseur @stevenhartley WDYT?

PLeVasseur (approved):
LGTM. Concept makes sense.

I suppose you're going to wait on a new release of `up-spec` with your upstreamed feature files and revise this PR, right?

sophokles73:
> LGTM. Concept makes sense.
> 
> I suppose you're going to wait on a new release of `up-spec` with your upstreamed feature files and revise this PR, right?

indeed

sophokles73:
Will be replaced with another PR implementing the Gherkin scenarios defined in up-spec v1.6.0-alpha.5

PR #256, opened 2025-05-22
Switch to use Eclipse Dash quevee action

AnotherDaniel:
The original quevee action has moved to the Eclipse Dash project last year; today, there was the first release/tag of the Dash qeuvee action, which going forward is going to be used to feed the Eclipse SDV Badge process. 
This PR switches our release pipeline to use the Eclipse Dash version of quevee.


PLeVasseur (approved):
Thanks for making the swap!

PR #275, opened 2025-08-01
added dynamic install of rust toolchain based on host arch and OS

Xerner:
The previous installation of the rust toolchain was done inside the `Dockerfile`. Docker has no knowledge of the host architecture and OS while it is building containers. Therefore, I moved the installation outside of the `Dockerfile` and into the a bash function that is called in the `post-create.sh` script. 

If using the Dev Containers extension in VS Code, the `post-create.sh` script is ran immediately after the container is created by VS Code, and installs the rust toolchain for your current hosts architecture and OS. It also adds the architecture and OS as a rust build target

For example, I am running WSL with an x86-64 architecture so it performed the following actions immediately after the Docker container built
- ran `rustup target add x86_64-unknown-linux-gnu`
- ran `rustup toolchain install stable-x86_64-unknown-linux-gnu`


.devcontainer/scripts/post-create.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+#region sudo stuff
+
+# This is a workaround which is againt necessary on MacOS 14.0, it looks like this bug is back:
```
AnotherDaniel, line 5: I haven't looked into details at all, and didn't test-run this myself, yet, but: if we now know our $ARCH_AND_OS, might it make sense to only run this workaround section when on MacOS? Could that be done? ...
I'll be able to actually test maybe later today, latest early next week.

AnotherDaniel (commented):
Good stuff, thank you!

I've added one thought wrt the MacOS workaround that I've originally added...

.devcontainer/scripts/post-create.sh
Xerner: I updated the script and wrapped the Mac workaround stuff with an OS check. I'm not 100% sure the check works as intended though. I've never used a Mac and don't own one

PR #277, opened 2025-09-25
Validate filter URIs used for registering listeners

sophokles73:
Improved logic for validating authority names in URIs to be more strict
and included validation of pattern URIs in checks for registering
listeners.


.github/workflows/check.yaml
```diff
@@ -135,7 +135,7 @@ jobs:
       - name: Run lychee
         uses: lycheeverse/lychee-action@v2
         with:
-          args: "--cache --max-cache-age 1d --verbose --no-progress --exclude-path './target/' --exclude-path './up-spec/' -- './**/*.md' './**/*.rs' './**/*.adoc'"
```
AnotherDaniel, line 138: does lychee have an issue with asciidoc?

AnotherDaniel (approved):
LGTM - a bit torn whether it'd be convenient or counterproductive to have Uuri constructors automatically lower-case any authority strings... but probably counterproductive, as we want people to be aware that they're using invalid names.

PR #279, opened 2025-09-29
Fix message expiration check

### LINKED ISSUE #278: UAttributes::check_expired panics for messages created in the future
The [UAttributes::check_expired](https://github.com/eclipse-uprotocol/up-rust/blob/7948332be4b025f9613dbf70c57fafb20292c2a0/src/uattributes.rs#L155) function panics if applied to attributes that have a message ID with a creation timestamp that lies in the future relative to the current system time.

This bug occurs when a another uEntity sends a message but its system clock is not in sync with the local system's clock.

sophokles73:
The message expiration check has been updated to properly consider
creation timestamps that lie in the future relative to the
local system time.

Fixes #278

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1526,7 +1526,7 @@ checksum = "1fc81956842c57dac11422a97c3b8195a1ff727f06e85c84ed2e8aa277c9a0fd"
 
 [[package]]
 name = "up-rust"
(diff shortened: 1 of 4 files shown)


AnotherDaniel (approved):
LGTM

PR #281, opened 2025-10-21
Add convenience accessors to UAttributes

sophokles73:
The accessors present in UMessage have been added to UAttributes
as well because some L2 API passes UAttributes around without
the full UMessage.


AnotherDaniel (approved):
Looking good. The name type_() feels a bit jarring to read, but it appears we've had that some time already on UMessage - and UMessage::message_type() does also feel redundant. So I guess I can't really come up with anything better...

Issue #282, opened 2025-10-23
Prototype how to use badges to signal up-spec level compatibility


PLeVasseur:
# Goal

Signal, at a glance, the level of [up-spec] compatibility that an Eclipse uProtocol project has using a badge present at the top of the repo `README.md`.

### Look and feel requirements

- LF.1: Badge should be colored either red or green
- LF.2: Badge that's red indicates the project complies with a non-current version of [up-spec]
- LF.3: Badge that's green indicates the projects complies with the currently released version of [up-spec]
- LF.4: Badge contains a string embedded in it indicating which version of [up-spec] is complied with of the form: `up-spec vx.y.z-a.b`
  - `x`: major version, numeric
  - `y`: minor version, numeric
  - `z`: patch version, numeric
  - `a`: one of `alpha`, `beta`, ...
  - `b`: `alpha`, `beta`, ... release number, numeric

### Functionality requirements
[Functionality requirements]: #functionality-requirements

- F.1: Clicking on badge should jump to GitHub tag on [up-spec] corresponding to that version
- F.2: Badge status should be checked and updated upon each PR merged to main
- F.3: If badge status has gone from being latest-compliant to non-latest-compliant, badge color should be changed from green to red
- F.4: If badge status has gone from being non-latest-compliant to latest-compliant, badge color should be changed from red to green
- F.5: If the version of [up-spec] complied with in the project is updated, the string embedded in badge must also be updated to match

### Project compliance requirements

- C.1: Badge is placed at the top of project `README.md`, right below project name

## Language libraries

Signal, at a glance, the level of [up-spec] compatibility that a language library has using a badge present at the top of repo `README.md`s.

###  Background

Language libraries, such as [up-rust], tend to incorporate [up-spec] as Git submodules
- to refer to the [up-core-api] folder which is then used to generate Protocol Buffers language-specific definitions
- access Gherkin files to perform testing, such as [uuid_protobuf_serialization.feature].

### Technical approach

The following needs investigation, but the idea is as follows:
- since language libraries use Git submodules to bring over [up-spec], perhaps we can specify [up-spec] based on its released tag
- then, implementing the Look and feel requirements' LF.4 becomes simpler since it may be possible to refer to the tag used for the Git submodule
- it may still be necessary to potentially use GitHub API(s) to query for tags over on [up-spec] to check for latest to implement Functionality requirements' F.3 and F.4

[up-spec]: https://github.com/eclipse-uprotocol/up-spec/
[...]

PR #283, opened 2025-10-24
added spec version and compliance check badge

Xerner:
This PR adds a workflow that adds badges that show 

- whether the current commit is on the latest up-spec version
- whether the current commit is up-spec compliant. This makes use of the existing requirement-tracing workflow

image

The actual implementation and one-time steps necessary for this to work are located here https://github.com/eclipse-uprotocol/ci-cd/pull/23

addresses #282

### DIFF
--- a/.github/workflows/create-badges.yaml
+++ b/.github/workflows/create-badges.yaml
@@ -36,17 +36,17 @@ jobs:
     steps:
       - name: check up-spec version
         id: check-up-spec-version
-        uses: xerner/ci-cd/.github/actions/check-spec-version@main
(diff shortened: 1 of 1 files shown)


Xerner:
Note: this PR was previously named "added spec compliance check badge" but actually just added a check for the latest spec version. So, I just added an additional badge that checks up-spec compliance and renamed the PR accordingly. Might as well have both, right?

image

I switched my implementation from using a Github Gist to directly pushing a commit to the repo from the Github Action run.

The PR is failing because the commit is authored by `action@github.com`, which is not an author covered by the Eclipse Fdn legal agreements. How should I go about addressing this? @sophokles73

PR #289, opened 2025-10-29
Make helper function public for external use.

sophokles73:
Made the UUri::verify_authority method public so that external
code can utilize it for URI authority verification.

This is particularly useful for transport implementations that are
created with a local authority to fall back to when source UUris
without authority are being used by client code.


AnotherDaniel (approved):
LGTM!

PR #293, opened 2025-11-14
Refactor Symphony unit tests and add example for Symphony Target

sophokles73:
The unit tests for the Symphony module have been refactored to remove
the dependency on the util feature when executing the tests, which has caused the nightly build to fail.

The original test using the LocalTransport has been refactored into an
example.


AnotherDaniel (approved):
LGTM

PR #294, opened 2025-11-20
Feat/up spec badge

senseix21:
This pull request introduces automation to keep the `up-spec` compatibility badge in the `README.md` up to date. A new GitHub Actions workflow checks which `up-spec` tag the project is using, compares it to the latest available tag, and updates the badge and its color accordingly in the `README.md`. This ensures the badge always reflects the current compatibility status with `up-spec`.

**Automation and badge update:**

* Added a new GitHub Actions workflow (`.github/workflows/up-spec-badge.yml`) that:
  * Determines the current and latest `up-spec` tags.
  * Decides the badge color (green if up to date, red otherwise).
  * Updates the badge in the `README.md` and commits the change automatically.

**Documentation:**

* Inserted an auto-updating `up-spec` compatibility badge section into the `README.md`, showing the current tag and its status.

### DIFF
--- a/.github/workflows/up-spec-badge.yml
+++ b/.github/workflows/up-spec-badge.yml
@@ -10,6 +10,7 @@ permissions:
 
 jobs:
   badge:
+    if: github.event.head_commit.author.name != 'github-actions[bot]'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout (with submodules)


PLeVasseur:
Hi @senseix21 👋 did you see there was already an existing PR here?
https://github.com/eclipse-uprotocol/up-rust/pull/283

(note that I've not looked into either PR)

Did this PR solve something notably different from the other one?

.github/workflows/up-spec-badge.yml
```diff
@@ -0,0 +1,95 @@
+name: Update up-spec compatibility badge
+
+on:
+  push:
+    branches: [ main ]
```
sophokles73: My understanding is that the whole mechanism will (currently) only work when we make changes to the main branch. However, IMHO we will need to be able to do this on (version specific) branches as well. For example once we have released stable versions 1.0.0 and 2.0.0 which implement different versions of the spec, we will need to be able to apply this mechanism on version specific bug-fix branches. AFAICT this mechanism should generally also work in such a scenario, but will need to consider the branch that we are pushing to in several places, or am I mistaken?

README.md
```diff
@@ -1,5 +1,10 @@
 # Eclipse uProtocol Rust library
 
+<!-- up-spec-badge-start -->
+[![up-spec](https://img.shields.io/badge/up--spec-v1.6.0--alpha.7-green)](https://github.com/eclipse-uprotocol/up-spec/tree/v1.6.0-alpha.7)
+<!-- up-spec-badge-end -->
+
+
```
sophokles73: It would be great if we could also update the link in the subsequent line(s) dynamically, based on the current tag ...

sophokles73 (commented):
Thanks for the PR, this looks promising :+1:
I am not too familiar with the use of the github-actions[bot] user. Will the commit created by this job be included in the list of commits from the (original) push that triggered the workflow in the first place?

PR #328, opened 2026-05-29
Remove protobuf generated types from public API

sophokles73:
Replaced all types that are generated by the protobuf compiler with custom types that do not depend on the protobuf crate. This reduces the public API surface and avoids leaking implementation details.

Added (Try)From implementations to convert between the protobuf generated types and the custom types.

Introduced finer grained feature gates for enabling generic protobuf support in L1 and L2 APIs and the default implementations of the L2 API.

This PR changes the default features to none, i.e. all functionality that goes beyond the L1 API needs to be enabled explicitly. Please refer to the Rust Docs for details.


src/lib.rs
```diff
@@ -102,12 +104,102 @@ pub use utransport::{MockLocalUriProvider, MockTransport, MockUListener};
...
+
+impl std::fmt::Display for SerializationError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+impl std::error::Error for SerializationError {}
+
+pub trait ProtobufMappable: Sized {
```
AnotherDaniel: Is this the best location for this code - considering especially that we might add other supported serialization formats in the future? Or am I prematurely optimizing with this question?

AnotherDaniel (approved):
This is is a massive change, kudos for pulling this off. I think the results are definitely worthwhile; I might have more input once adopting this for usubscription, until then looks good to me!

src/lib.rs
```diff
@@ -102,12 +104,102 @@ pub use utransport::{MockLocalUriProvider, MockTransport, MockUListener};
 mod uuid;
 pub use uuid::UUID;
 
-// protoc-generated stubs, see build.rs
-mod up_core_api {
+// protoc-generated types, see build.rs
+pub mod up_core_api {
```
PLeVasseur: This makes the full generated protobuf module public as `up_rust::up_core_api::*`. I understand that some lower-level integrations and external crates may still need access to generated protobuf types, so I do not think this PR necessarily has to remove protobuf from every public surface.

Could we make the boundary explicit, though? For example:

- keep handwritten wrapper types as the default ergonomic SDK API;
- expose generated protobuf types only through a clearly documented low-level/wire module or feature, such as `proto-api`;
- document that this low-level API exists for interoperability and is not the preferred stable ergonomic API;
- avoid relying on generated types from examples/tests unless those tests are specifically exercising protobuf interoperability.

That would preserve the escape hatch for crates that still need generated contracts while keeping the PR's stated API-boundary goal clear.

src/ustatus.rs
```diff
@@ -115,116 +210,158 @@ impl UStatus {
...
     }
 }
 
 impl Error for UStatus {}
 
+impl From<UStatusProto> for UStatus {
+    fn from(value: UStatusProto) -> Self {
+        let message = value.message.unwrap_or_default();
+        UStatus {
+            code: UCode::from(value.code.enum_value_or_default()),
```
PLeVasseur: `enum_value_or_default()` turns unknown protobuf enum values into the enum default, which is `OK`. That means invalid or newer wire data like `code = 12345` becomes a successful `UStatus`.

`up-spec` defines the valid `UCode` values and `OK = 0`; the issue here is the Rust wrapper's conversion policy for unrecognized wire values. Since `UStatus` cannot preserve unknown enum values, this conversion should be fallible rather than defaulting.

Can we change this path to reject unknown values, for example by using `value.code.enum_value()` or `UCode::from_i32(value.code.value())`, and returning a conversion/serialization error for unrecognized values while still mapping explicit protobuf `UNKNOWN` to `UCode::Unknown`?

It may also be worth a follow-up `up-spec` clarification that SDKs must not coerce unknown enum values to proto defaults when converting into strongly typed wrappers.

src/ustatus.rs
```diff
@@ -115,116 +210,158 @@ impl UStatus {
...
+        }
+    }
+}
+
+impl From<&UStatus> for UStatusProto {
+    fn from(value: &UStatus) -> Self {
+        UStatusProto {
+            code: UCodeProto::from(value.code).into(),
+            message: Some(value.message.as_str().to_string()),
+            ..Default::default()
```
PLeVasseur: The wrapper currently loses spec-modeled `UStatus` data on protobuf/custom/protobuf round trip. `up-spec` defines `UStatus` with `code`, optional `message`, and repeated `details`; the handwritten wrapper only stores `code` and a non-optional `String` message.

As a result:

- `details` is always dropped;
- absent `message` becomes `""`;
- serializing back always writes `Some("")` for an absent message.

Can we either preserve these fields in the wrapper, or explicitly document/test that this wrapper is intentionally lossy? Since `details` uses protobuf `Any`, a small handwritten wrapper such as `UAny { type_url: String, value: Vec<u8> }` would preserve the data without exposing generated protobuf types directly.

src/communication/udiscovery_client.rs
```diff
@@ -67,10 +84,13 @@ impl UDiscovery for RpcClientUDiscovery {
...
             .map(|response_message| {
-                response_message
-                    .uris
-                    .as_ref()
-                    .map_or(vec![], |batch| batch.uris.to_owned())
+                response_message.uris.as_ref().map_or(vec![], |batch| {
+                    batch
+                        .uris
+                        .iter()
+                        .filter_map(|uri| UUri::try_from(uri).ok())
```
PLeVasseur: This silently drops malformed URIs from a successful uDiscovery response. The service contract returns a batch of `UUri` values; if any returned URI cannot be represented as a valid Rust `UUri`, the client should not turn that into an undetectable partial success.

Can we collect the conversions as a `Result<Vec<_>, _>` and return an error on the first malformed URI? If partial results are desired, the API should expose that explicitly with diagnostics rather than silently filtering entries.

This may also be worth a follow-up spec/TCK clarification: client libraries should not silently discard malformed service-response entries when adapting typed service contracts.

src/communication/udiscovery_client.rs
```diff
@@ -92,7 +112,13 @@ impl UDiscovery for RpcClientUDiscovery {
                 request_message,
             )
             .await
-            .map(|response_message| response_message.topics.to_owned())
+            .map(|response_message| {
+                response_message
+                    .topics
+                    .iter()
+                    .filter_map(|topic| TopicInfo::try_from(topic).ok())
```
PLeVasseur: This `filter_map` turns malformed `ServiceTopicInfo` entries into a successful partial result. `TopicInfo::try_from` already returns `UStatus` for missing or malformed topic data and missing `info`, but the error is discarded here.

Can we propagate that conversion error instead of dropping the entry?

```rust
response_message
    .topics
    .iter()
    .map(TopicInfo::try_from)
    .collect::<Result<Vec<_>, _>>()
```

If partial results are intended, the API should probably make that explicit and expose diagnostics. Otherwise malformed uDiscovery responses become very hard to diagnose.

This might also be a reasonable follow-up area for `up-spec`/TCK clarification: typed SDKs should not silently filter malformed service-response entries.

src/core/usubscription.rs
```diff
@@ -163,93 +280,123 @@ pub fn usubscription_uri(resource_id: u16) -> UUri {
...
+    ///   This attribute is commonly used for mobile/cloud components subscribing to vehicle topics that are published
+    ///   at a high rate. If the desired sampling period set by the subscriber is lower than the original publisher's
+    ///   publication period, the attribute is ignored.
+    ///   If not specified, the sampling period is set by the publisher.
     ///
     /// # Returns
     ///
-    /// * [`SubscriptionResponse`] detailing if subscription was successful with other metadata
+    /// * The outcome of the attempt to establish the subscription.
     async fn subscribe(
```
PLeVasseur, line 201: This narrows the public uSubscription API from the service contract to `topic`, `expiration`, and `min_sample_period`, returning only `SubscriptionStatus`. That covers the common path, but it drops several modeled request/response fields from the API surface:

- request `SubscribeAttributes.details` cannot be sent;
- response `SubscriptionResponse.config` cannot be returned, even though the proto comments say it may contain event delivery configuration required to consume events;
- response `SubscriptionResponse.topic` is not exposed;
- `SubscriptionStatus.message` is discarded by the wrapper.

Could we keep API coverage without exposing generated protobuf types? One option is to keep this convenience method but add a complete handwritten `SubscribeOptions` / `SubscribeResponse` path that carries attributes, response topic, status message, and delivery config. If this trait intentionally exposes only the simple subscription flow, please document that limitation explicitly.

src/uattributes.rs
```diff
@@ -595,15 +661,148 @@ impl UAttributes {
...
+impl TryFrom<&UAttributesProto> for UAttributes {
+    type Error = UAttributesError;
+
+    fn try_from(value: &UAttributesProto) -> Result<Self, Self::Error> {
+        use crate::up_core_api::uattributes::UPriority as UPriorityProto;
+
+        Ok(UAttributes {
+            commstatus: value
+                .commstatus
+                .map(|cs| UCode::from(cs.enum_value_or_default())),
```
PLeVasseur: This has the same invalid-enum issue as `UStatus`: `enum_value_or_default()` maps an unknown protobuf `commstatus` value to `OK`.

`commstatus` is used for delivery errors on RPC Request or Response messages, so an invalid or unrecognized communication status should not become success. Since the wrapper cannot preserve unknown enum values, this conversion should be fallible:

```rust
commstatus: match value.commstatus {
    Some(cs) => Some(UCode::from_i32(cs.value()).ok_or_else(|| {
        UAttributesError::validation_error(format!("unknown commstatus value: {}", cs.value()))
    })?),
    None => None,
}
```

A regression test with generated `UAttributes` containing `commstatus = EnumOrUnknown::from_i32(12345)` would ensure `UAttributes::try_from` or `UMessage::try_from` fails instead of producing `Some(UCode::Ok)`.

This may also be worth a follow-up spec/SDK clarification: unknown enum values must not be coerced to protobuf defaults in strongly typed wrappers.

src/core/usubscription.rs
```diff
@@ -136,6 +45,213 @@ pub const RESOURCE_ID_RESET: u16 = 0x0009;
...
+pub enum SubscriptionStatus {
+    Unsubscribed = State::UNSUBSCRIBED as isize,
+    SubscribePending = State::SUBSCRIBE_PENDING as isize,
+    Subscribed = State::SUBSCRIBED as isize,
+    UnsubscribePending = State::UNSUBSCRIBE_PENDING as isize,
+}
+
+impl From<&SubscriptionStatusProto> for SubscriptionStatus {
+    fn from(status: &SubscriptionStatusProto) -> Self {
+        let state = status.state.enum_value_or_default();
```
PLeVasseur: `enum_value_or_default()` maps an unknown subscription state to protobuf default `UNSUBSCRIBED`. That means invalid service data can become a real `SubscriptionStatus::Unsubscribed`.

Could we make this conversion fallible:

- replace `From<&SubscriptionStatusProto>` with `TryFrom<&SubscriptionStatusProto>`;
- use `status.state.enum_value()` or `State::from_i32(status.state.value())`;
- return an error for an unknown state and propagate that through `subscribe`, `SubscriptionInfo::try_from`, and update parsing.

`UCode::Internal` seems appropriate because this is malformed service response data, but the exact status code is not currently specified by `up-spec`. If we want consistent behavior across SDKs, this would be a good follow-up clarification for invalid service responses.

We could consider a test that constructs a generated subscription status with an unknown state value and asserts that conversion fails.

src/core/usubscription.rs
```diff
@@ -136,6 +45,213 @@ pub const RESOURCE_ID_RESET: u16 = 0x0009;
...
+            .ok_or(UStatus::fail_with_code(
+                UCode::InvalidArgument,
+                "status missing",
+            ))?;
+        subscription
+            .attributes
+            .as_ref()
+            .ok_or_else(|| UStatus::fail_with_code(UCode::InvalidArgument, "missing attributes"))
+            .map(|attributes| {
+                let expiration = attributes.expire.as_ref().map(|ts| ts.seconds as u64);
```
PLeVasseur: This casts protobuf timestamp seconds to `u64` with `as`, so negative service timestamps wrap into huge future expiration values.

```rust
let expiration = attributes.expire.as_ref().map(|ts| ts.seconds as u64);
```

Could we validate the timestamp instead:

- use `u64::try_from(ts.seconds)` and return an error for negative seconds;
- decide whether non-zero `nanos` should be preserved, rejected, or explicitly truncated/documented by this seconds-only wrapper;
- consider a small helper for converting protobuf `Timestamp` into the wrapper's public expiration representation.

This appears in both `SubscriptionInfo::try_from(Subscription)` and `SubscriptionInfo::try_from(Update)`.

This may also be worth a follow-up `up-spec`/SDK clarification: if an SDK exposes subscription expiration as epoch seconds rather than full protobuf `Timestamp`, it should define required behavior for negative seconds, non-zero nanos, and malformed service-response timestamps.

src/cloudevents.rs
```diff
@@ -176,7 +176,7 @@ impl CloudEvent {
     }
 
     fn set_commstatus(&mut self, status: UCode) {
-        if status != UCode::OK {
+        if status != UCode::Ok {
```
PLeVasseur, line 241: `get_commstatus()` silently ignores invalid `commstatus` values with `and_then(UCode::from_i32)`. For a response CloudEvent, `commstatus = 12345` becomes `None`, and absence is later treated like no communication status rather than invalid input.

Can this path fail when the extension is present but cannot be decoded as a known `UCode`? Ideally it should also reject a `commstatus` extension with the wrong CloudEvent attribute type instead of treating it like integer `0` or absence.

A regression test could build a response CloudEvent with integer `commstatus = 12345` and assert conversion fails instead of producing `commstatus() == None`.

This is also a good candidate for an `up-spec` clarification: receivers should reject present-but-invalid CloudEvent extension values for enum-backed uProtocol attributes rather than treating them as absent/default.

src/communication/usubscription_client.rs
```diff
@@ -52,23 +55,81 @@ impl RpcClientUSubscription {
...
+        min_sample_period: Option<u32>,
+    ) -> Result<SubscriptionStatus, UStatus> {
+        let subscription_request = SubscriptionRequest {
+            topic: Some(topic.into()).into(),
+            attributes: match (expiration, min_sample_period) {
+                (None, None) => None.into(),
+                _ => Some(SubscribeAttributes {
+                    expire: expiration
+                        .map(|ts| Timestamp {
+                            seconds: ts as i64,
```
PLeVasseur: `expiration` and `before` are `u64`, but they are converted to protobuf `Timestamp.seconds` with `as i64`. Values greater than `i64::MAX` wrap to negative seconds before being sent to uSubscription.

Could we validate at the API boundary? That would help avoid sending a wrapped value the caller did not intend. A small helper using `i64::try_from` could return `UCode::InvalidArgument` when the caller provides a value that cannot be represented safely as a protobuf timestamp. If we want full protobuf `Timestamp` validation, the helper can also enforce the protobuf seconds range and set `nanos = 0` explicitly.

If we go that route, then the same checked conversion should be used for `subscribe(expiration)` and `reset(before)`.

src/umessage.rs
```diff
@@ -72,107 +82,104 @@ impl From<&str> for UMessageError {
...
-    /// Gets this message's attributes.
-    pub fn attributes_unchecked(&self) -> &UAttributes {
-        self.attributes().expect("message has no attributes")
+    pub(crate) fn new(
+        attributes: UAttributes,
+        payload: Option<Bytes>,
+    ) -> Result<Self, UMessageError> {
+        Ok(UMessage {
+            attributes,
+            payload: payload.map(|p| p.to_vec()),
```
PLeVasseur: This changes payload ownership from `Bytes` to `Vec<u8>`. `UMessage::new` copies any `Some(Bytes)` payload with `to_vec()`, and `UMessage::clone()` now deep-copies payload bytes.

I put together a small benchmark comparing `main` and this PR: https://github.com/PLeVasseur/up-rust/tree/pr-328-payload-clone-benchmark

In that benchmark, clone cost on `main` stays roughly flat at about 80-90 ns regardless of payload size because generated `UMessage.payload` is `Option<Bytes>`. On this PR, empty/tiny messages are faster, but payload-bearing clone cost scales with payload size: about 21.8 us/clone for 1 MiB and about 88.2 us/clone for 4 MiB in the benchmark run.

If the goal is to hide protobuf-generated types rather than change payload ownership, could `UMessage` keep `Bytes` internally while exposing `payload() -> Option<&[u8]>`? That preserves the new public API shape and also allows wrapper/protobuf conversions to clone `Bytes` cheaply instead of copying from `Vec<u8>` back into `Bytes`.

PLeVasseur (changes requested):
Thanks for taking this on. The direction is good: moving the ergonomic SDK API away from generated protobuf structs is the right boundary.

I think there are a few places worth tightening before merge. The main API-surface gap is that generated protobuf APIs are still publicly reachable via `up_rust::up_core_api::*`, and `core::utwin` still re-exports generated response types behind the `utwin` feature. If keeping generated protobuf types reachable is necessary for lower-level integrations or for crates that still consume the generated wire contracts directly, could we make that an explicit API boundary? For example, keep the ergonomic API surface protobuf-free where wrappers exist, but expose generated types through a clearly named/documented low-level module or feature such as `proto-api`/`wire-api`.

That would also give us a concrete path for remaining generated re-exports like `core::utwin`: either wrap them now, or document them as part of the low-level protobuf escape hatch until handwritten wrappers exist. The important thing is that callers can tell which API is intended to be stable and ergonomic versus which API is generated/wire-level compatibility surface.

I also found several conversion paths where wire/service data can be silently changed or dropped:

- unknown protobuf enum values can become real default states like `OK` or `UNSUBSCRIBED`;
- unknown CloudEvent `commstatus` values are ignored instead of rejected;
- `UStatus.details`, uSubscription status messages, request `SubscribeAttributes.details`, and response `SubscriptionResponse.config`/`topic` are not represented in the wrappers;
- malformed uDiscovery results are filtered into successful partial results;
- timestamp conversions use unchecked `as` casts in both directions.

Most of these look fixable with fallible conversions (`TryFrom` instead of `From` where wire data can be invalid), small handwritten wrapper/options types for structured fields, collecting conversion results instead of `filter_map`-ing malformed service responses, and checked integer conversions for timestamps.

Since this PR intentionally changes a lot of public API shape, it would also help to add a short migration note in the PR or release notes for callers moving from generated structs to the new wrappers.

src/core/usubscription.rs
AnotherDaniel: One note on this: many of these semi-defined details of the uSubscription spec might be dropped soon - I'm working on that PR where we clean up usub spec between adoc, asyncapi and remaining proto message definitions, and currently I'm partial to dropping all these fields that are only very vaguely described, and where actual usage details or examples are missing. 
Also tbd, of course - but the delivery config is one of those fields.

src/lib.rs
sophokles73: I have changed the module's visibility to `pub(crate)`. IMHO client code should not really need access to any of its types because all the public APIs at L1 and L2 now only depend on public types not created by protoc. We should now only need the protoc types for mapping protobufs, but that is an internal implementation detail.

src/cloudevents.rs
sophokles73: You are right, actually most of the getters currently run into this problem.
I will change them accordingly so that they return `Result<Option<T>>, UAttributesError>` instead of just `Option<T>` ...

sophokles73:
@PLeVasseur I have (hopefully) addressed all the issues that you had raised in the review. Would you mind taking another look?

@AnotherDaniel can you also please take another look?

src/uattributes.rs
```diff
@@ -595,15 +659,155 @@ impl UAttributes {
...
+    }
+
+    impl TryFrom<&UAttributesProto> for UAttributes {
+        type Error = UAttributesError;
+
+        fn try_from(attribs_proto: &UAttributesProto) -> Result<Self, Self::Error> {
+            Ok(UAttributes {
+                commstatus: attribs_proto
+                    .commstatus
+                    .map(|cs| UCode::from(cs.enum_value_or_default())),
```
PLeVasseur: Thanks for fixing this same class of issue in `UStatus`, subscription state conversion, and CloudEvents. One protobuf path still appears to use `enum_value_or_default()` for `UAttributes.commstatus`, so an unknown wire value like `commstatus = 12345` can still become `Some(UCode::Ok)`.

Could we make this match the other fallible enum conversions by using `cs.enum_value().map(UCode::from)` and mapping the error to `UAttributesError::validation_error(...)`? A small `UAttributes::try_from` regression test with `EnumOrUnknown::from_i32(12345)` would lock this down without needing to exercise the broader `UMessage` path.

src/core/usubscription/usubscription_client.rs
```diff
@@ -14,21 +14,189 @@
...
+    up_core_api::usubscription::{
+        FetchSubscribersResponse, FetchSubscriptionsRequest, FetchSubscriptionsResponse,
+        NotificationsResponse, ResetResponse, SubscribeAttributes, Subscription,
+        SubscriptionRequest, SubscriptionResponse, UnsubscribeRequest, UnsubscribeResponse, Update,
+    },
+    UCode, UStatus, UUri,
 };
 
-use super::{CallOptions, RpcClient};
+fn unix_epoch_millis_as_protobuf_timestamp(
```
PLeVasseur, line 35: Thanks for switching this away from unchecked `as i64`; the overflow case is much safer now. The main remaining issue is unit consistency: the public `USubscription` trait docs still say `expiration` and `before` are seconds since Unix epoch, while this helper and implementation treat them as milliseconds.

Could we choose one unit at the public API boundary and make the trait docs, helper name, request conversion, and `SubscriptionInfo::expiration()` docs match? Either keep the public API as seconds and set `Timestamp { seconds, nanos: 0 }`, or explicitly document/rename the public API as milliseconds.

While touching the reverse conversion, can we also validate `0 <= Timestamp.nanos < 1_000_000_000` before converting it to milliseconds? Today `ts.nanos as u64` can turn negative nanos into a huge positive offset, and out-of-range positive nanos are accepted/truncated. Focused tests for `subscribe`/`reset` unit conversion and `SubscriptionInfo::try_from` with invalid nanos would cover this boundary.

src/core/usubscription.rs
```diff
@@ -163,93 +156,123 @@ pub fn usubscription_uri(resource_id: u16) -> UUri {
...
+    ///   This attribute is commonly used for mobile/cloud components subscribing to vehicle topics that are published
+    ///   at a high rate. If the desired sampling period set by the subscriber is lower than the original publisher's
+    ///   publication period, the attribute is ignored.
+    ///   If not specified, the sampling period is set by the publisher.
     ///
     /// # Returns
     ///
-    /// * [`SubscriptionResponse`] detailing if subscription was successful with other metadata
+    /// * The outcome of the attempt to establish the subscription.
     async fn subscribe(
```
PLeVasseur, line 201: Given the spec-cleanup context, I do not think this has to block on wrapping every modeled field immediately. Non-blocking docs/API-contract ask: could we document `USubscription::subscribe` as the simple/common subscription path?

For example, a short note could say that this convenience API intentionally exposes topic, expiration, minimum sample period, and subscription state, and currently does not expose protobuf `SubscribeAttributes.details`, status message, response topic, or event delivery config. If those fields survive the spec cleanup, we can add a richer handwritten options/response API later without exposing generated protobuf types.

Cargo.toml
```diff
@@ -33,22 +33,29 @@ rust-version = "1.88"
 version = "0.10.0-SNAPSHOT"
 
 [features]
-default = ["communication"]
-cloudevents = []
-communication = ["usubscription", "dep:thiserror", "tokio/sync", "tokio/time"]
-udiscovery = []
-usubscription = []
-utwin = []
+default = []
```
PLeVasseur, line 36: This default-feature change is probably the right direction for a smaller dependency surface, but it is also a migration hazard for existing callers: `up-rust = "0.9"` used to expose the communication stack by default, while this PR requires callers to opt into `communication` or the narrower L2 features.

Could we add a short migration note, either in the README, PR description, or release notes? I think a brief paragraph is enough: default features are now empty; callers wanting the old default behavior should enable `features = ["communication"]`; protobuf helpers are feature-gated by `protobuf-support` / `up-core-types`; generated `up_core_api` is no longer public; `utwin` is removed; and `UStatus` message/details access changed.

That would make the intentional API boundary change much easier for downstream users to apply.

PLeVasseur (changes requested):
Great stuff! I see that you incorporated the comments.

I ended up finding a few more spots that'd be good to address.

Couple of them are a little closer to nits, but couple of them are areas we could see further improvement.

src/uattributes.rs
sophokles73: Ok, I hope that I have covered all cases now where the mapping of attributes could go wrong. I have also created a unit test that verifies the mapping functionality.

src/core/usubscription/usubscription_client.rs
sophokles73: The USubscription trait now uniformly uses _milliseconds since Unix epoch_ for specifying instants.
I have also added a corresponding check for the Timestamp.nanos value to be within range and have added unit tests.

src/core/usubscription.rs
sophokles73: Sorry, I do not get your point. Can you elaborate what is missing or is wrong with the proposed function definition?

Cargo.toml
sophokles73: FMPOV it should be sufficient to include this in the Release Notes (on GitHub). That is where people will look (after having found out that we do not maintain an ongoing changelog within the code base).

I have added corresponding wording to the PR description and added the _breaking change_ label to the PR.

WDYT?

sophokles73:
@PLeVasseur I have pushed more changes ...

BTW would you mind resolving the comments that have been (sufficiently) addressed?

Cargo.toml
PLeVasseur: Yep, that sounds reasonable to me. Thanks for this.

src/core/usubscription/usubscription_client.rs
PLeVasseur: Thanks, the conversion helper, `nanos` validation, and tests look much safer now.

One small public-doc cleanup still seems to remain: this wrapper now consistently treats subscription instants as milliseconds, but `USubscription::subscribe` still describes `expiration` as "seconds since Unix epoch". Could we update that one line to milliseconds as well? After that, this thread looks good to resolve from my side.

src/core/usubscription.rs
PLeVasseur: Yep, thanks for asking; I should have been clearer.

My question is only about the intended contract of this method. Compared to the current v3 protobuf `Subscribe()` contract, returning only `SubscriptionStatus` is narrower. If full v3 contract coverage were the goal, I think users would expect a handwritten Rust type for [`SubscriptionResponse`](https://github.com/eclipse-uprotocol/up-spec/blob/94b76ad3b8083d4bb0e7d22305c8de2677051a4a/up-core-api/uprotocol/core/usubscription/v3/usubscription.proto#L188).

Given the uSubscription cleanup that @AnotherDaniel plans to do in v4 that may narrow some of the current protobuf fields anyway, I am not asking for that richer API in this PR. Maybe we could more simply document this method as the convenience/common-path API?

PLeVasseur (changes requested):
Okay, tried to clarify my thoughts a bit more on the two remaining things I saw.

Hopefully that did the trick, but if not, please lemme know.

src/core/usubscription/usubscription_client.rs
sophokles73: Oh, I thought that I had also committed that change already. I'll fix it. Thanks for spotting.
In general, given that @AnotherDaniel is currently reworking that whole service, I guess these functions will need to be adapted soon enough again ;-)

src/core/usubscription.rs
sophokles73: Ok, I think I see where you're heading. @AnotherDaniel will create V4 of the uSubscription service definition and can therefore simply break the existing API and simplify, remove params etc.
This PR, however, is still based on V3 of uSubscription and should therefore support the parameters that have been defined in the V3 service spec, is that your point, @PLeVasseur?
I guess it all depends on whether V4 should actually _replace_ V3 or should simply be added to the repo. IMHO we should simply replace V3 because, frankly, nobody had been using it anyway so far.
We should then only provide a client based on the V4 contract and simply leave the V3 contract be.

However, if we are doing another up-rust release after this PR has been merged, that would still be based on (i.e. implement) v1.6.0-alpha.7 and thus should implement the V3 uSubscription contract.
I guess I will have to bite the bullet and (until we create the next version of up-rust with only uSubscription V4 in it), keep the uSubscription client on the V3 level ...

@PLeVasseur @AnotherDaniel WDYT?

src/core/usubscription/usubscription_client.rs
PLeVasseur: > given that @AnotherDaniel is currently reworking that whole service, I guess these functions will need to be adapted soon enough again ;-)

=]

src/core/usubscription.rs
PLeVasseur: > This PR, however, is still based on V3 of uSubscription and should therefore support the parameters that have been defined in the V3 service spec, is that your point, @PLeVasseur?

Yep, exactly =]

> IMHO we should simply replace V3 because, frankly, nobody had been using it anyway so far.

Yeah, agreed. As soon as V4 is available, we should simply replace it. No need for deprecation warning and versions of compatibility or anything like that.

> I guess I will have to bite the bullet and (until we create the next version of up-rust with only uSubscription V4 in it), keep the uSubscription client on the V3 level ...

Yeah... or to document that this API is narrower for the time being, till V4 comes? Depends on if you think it will be valuable, given the (lack of) users of the APIs at the moment. 🤔

sophokles73: I have added some wording explaining the reduced set of supported input/output parameters to the trait's doc comment.

sophokles73:
@PLeVasseur I have pushed some (final?) changes. WDYT?

PLeVasseur (approved):
Hey @sophokles73 -- thanks for making these small remaining updates to clarify the current state of `USubscription`.

Let's :shipit:

src/lib.rs
sophokles73: I have moved it to its own module ...

PR #338, opened 2026-06-18
Set default permissions for GitHub workflows to "none"

sophokles73:
Only grant "write" permissions to the "contents" permission where necessary for uploading workflow artifacts. This is a security best practice to follow the principle of least privilege and minimize the potential attack surface of the workflows.


AnotherDaniel (approved):
Looks good - the Copilot remarks about read permissions might be worth looking into, though?

PR #340, opened 2026-06-19
Use Trusted Publishing for Release workflow

sophokles73:
Modified the release.yaml workflow to no longer use a long lived secret for authentication to crates.io, but instead use Trusted Publishing based on short lived tokens as described in https://crates.io/docs/trusted-publishing


AnotherDaniel (approved):
Looks good - I feel the same overhaul coming for up-subscription-rust...

sophokles73:
and all the other repos :-)

PR #341, opened 2026-06-22
Use Typestate pattern for UMessageBuilder

sophokles73:
Introduced the Typestate pattern for the UMessageBuilder to enforce the correct sequence of method calls when building a UMessage of a particular type.

This change ensures that all required (and only appropriate) fields are set before the message can be built, improving code safety and reducing the likelihood of runtime errors.

### DIFF
--- a/src/umessage/umessagebuilder.rs
+++ b/src/umessage/umessagebuilder.rs
@@ -377,7 +377,7 @@ impl UMessageBuilder<InitialBuilderState> {
             message_type: UMessageType::Response,
             source: request_attributes
                 .sink()
-                .expect("Request attributes must have a sink")
+                .expect("Request attributes must contain a sink")
                 .to_owned(),
             message_id: None,
             sink: Some(request_attributes.source().to_owned()),
@@ -781,10 +781,6 @@ impl UMessageBuilder<RequestBuilderState> {
     ///
     /// The builder.
     ///
-    /// # Panics
-    ///
-    /// * if the message is not an RPC request message
-    ///
     /// # Examples
     ///
     /// ```rust
@@ -816,11 +812,6 @@ impl UMessageBuilder<RequestBuilderState> {
     /// # Returns
     ///
     /// The builder.
-    ///
-    /// # Panics
-    ///
-    /// * if the given level is greater than [`i32::MAX`]
-    /// * if the message is not an RPC request message
     ///
     /// # Examples
     ///
@@ -854,10 +845,6 @@ impl UMessageBuilder<ResponseBuilderState> {
     /// # Returns
     ///
     /// The builder.
-    ///
-    /// # Panics
-    ///
-    /// * if the message is not an RPC response message
     ///
     /// # Examples
     ///
--- a/src/umessage/umessagebuilder.rs
+++ b/src/umessage/umessagebuilder.rs
@@ -21,28 +21,90 @@ use crate::{
     UCode, UMessage, UMessageError, UMessageType, UPayloadFormat, UPriority, UUri, UUID,
 };
 
-/// A builder for creating [`UMessage`]s.
+mod sealed {
+    pub trait Sealed {}
+}
+
+/// Represents the type of message being built by a [UMessageBuilder].
+/// This is used to limit the available builder functions and to ensure that the
+/// builder is used in a consistent way.
+/// For example, only a builder for a request message can set the `permission_level` and `token` attributes
+/// and only a builder for a response message can set the `comm_status` and `request_id` attributes.
+/// This is achieved by having different builder state types for each message type and by using the
+/// [BuilderState::merge_into_attributes] function to add the state-specific attributes to the
+/// final message attributes when building the message.
 ///
-/// Messages are being used by a uEntity to inform other entities about the occurrence of events
-/// and/or to invoke service operations provided by other entities.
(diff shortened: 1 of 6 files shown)


src/umessage/umessagebuilder.rs
```diff
@@ -21,28 +21,90 @@ use crate::{
...
 
-impl UMessageBuilder {
+/// A builder for creating [`UMessage`]s.
+///
+/// Messages are used by uEntities to inform other entities about the occurrence of events
+/// and/or to invoke service operations provided by other entities.
+///
+/// For each type of message there is a dedicated builder which ensures that only attributes
+/// relevant to that message type are set.
+pub struct UMessageBuilder<S: BuilderState> {
```
PLeVasseur, line 102: This makes the builder's state type part of the public API, but only `UMessageBuilder` itself is re-exported from the crate root. The state marker types and `BuilderState` trait are not publicly reachable.

That may be the right choice if we want to support only immediate fluent construction:

```rust
UMessageBuilder::publish(topic).build()?;
```

Could we make that choice explicit, though? If downstream crates are expected to compose with partially configured builders, then the unexported state types make a few reasonable patterns hard to express.

For example, a wrapper crate may want to centralize project defaults but still let the caller decide the payload:

```rust
fn telemetry_event(topic: UUri) -> UMessageBuilder<PublishBuilderState> {
    let mut builder = UMessageBuilder::publish(topic);
    builder.with_priority(UPriority::CS2);
    builder.with_traceparent(current_traceparent());
    builder
}
```

Or middleware/test utilities may want to apply common metadata to any builder state, similar to `apply_common_options` in this crate:

```rust
fn apply_observability<S: BuilderState>(builder: &mut UMessageBuilder<S>) {
    builder.with_traceparent(current_traceparent());
}
```

If those patterns are intentionally out of scope, it would be good to document that `UMessageBuilder` is intended for direct fluent use and not as a type callers should name in their own APIs. If they are in scope, could we expose that intentionally, perhaps via public aliases such as `PublishMessageBuilder`, `RequestMessageBuilder`, and `ResponseMessageBuilder` instead of requiring users to depend directly on the marker type names?

src/umessage/umessagebuilder.rs
```diff
@@ -808,6 +770,119 @@ impl UMessageBuilder {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `token` - The token.
+    ///
+    /// # Returns
+    ///
+    /// The builder.
+    ///
+    /// # Panics
```
PLeVasseur: These docs still describe the old runtime behavior: `with_token`, `with_permission_level`, and `with_comm_status` say they panic when used with the wrong message type.

`with_permission_level` also says it panics if the value is greater than `i32::MAX`, but the method takes `u32` and validation now reports inconsistent attributes as a `UMessageError` from `build()`.

With the typestate API, those methods are no longer available on the wrong builder state, so misuse fails at compile time instead of panicking. Could we update these sections to describe that compile-time restriction and update or remove the stale panic bullets? That will make the docs match the main value of this PR.

PLeVasseur (changes requested):
Thanks for putting this together. The typestate direction is a nice fit for the builder: it moves the request-only and response-only attributes out of runtime panics and into the type system, while still keeping the normal call sites readable.

I found two things worth tightening before merge.

- The new public `UMessageBuilder<S>` shape exposes the state type in the public API, but the state marker types and `BuilderState` trait are not reachable from the public crate API. Chained constructor usage works, but external users cannot conveniently name, store, or return a typed builder from their own APIs.
- The request/response-specific method docs still describe runtime panics for using the method on the wrong message type. With this PR those methods are simply not available for the wrong state, so the docs should be updated to describe the compile-time restriction instead.

src/umessage/umessagebuilder.rs
sophokles73: Great finding. IMHO we should not constrain ourselves and users to fluent invocation only. I have several cases in the code base where I also need to invoke builder functions in a non-fluent way. I will export the BuilderState traits at the lib.rs level to mitigate this problem. Or do you see an issue with doing that apart from the fact that users might need to pull these traits into scope in order to use the builder efficiently? Maybe we should start thinking about creating a prelude?

src/umessage/umessagebuilder.rs
sophokles73: Absolutely, I seem to have forgotten to remove these

sophokles73: I have removed the corresponding sections from the Rust doc comments.

sophokles73:
@PLeVasseur I have pushed corresponding changes. Would you mind taking another look?

src/umessage/umessagebuilder.rs
PLeVasseur: > Maybe we should start thinking about creating a prelude?

I see this was not done in this one, so maybe you're thinking to do this in a follow-up which can consider this more carefully, separately. Is that the case?

sophokles73: yep, that's the plan.

PLeVasseur: Gotcha; thx. Then let's :shipit:

PLeVasseur (approved):
LGTM, let's :shipit:

PR #342, opened 2026-06-23
Fix symphony_target example requirements

sophokles73:
Added "up-l2-rpc-client" and "up-l2-rpc-server" features to the symphony_target example requirements in Cargo.toml. This ensures that the example can be built and run correctly with all necessary dependencies.


AnotherDaniel (approved):
LGTM

PR #358, opened 2026-07-20
Usubscription v4

AnotherDaniel:
- implement all changes required to support usubscription spec v4
- some dependency version bumps

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,19 +59,29 @@ util = ["tokio/sync"]
 [dependencies]
 async-trait = { version = "0.1" }
 bytes = { version = "1.12.1" }
+chrono = { version = "0.4.45", features = ["now"] }
 mediatype = "0.23"
 mockall = { version = "0.15", optional = true }
 protobuf = { version = "3.7.2", features = ["with-bytes"], optional = true }
 rand = { version = "0.10.2" }
-regex = { version = "1.13.1", default-features = false, features = ["perf", "std"] }
+regex = { version = "1.13.1", default-features = false, features = [
+    "perf",
+    "std",
+] }
 serde = { version = "1.0.228", optional = true }
 serde_json = { version = "1.0.150", optional = true }
 symphony = { version = "0.1.5", optional = true }
 thiserror = { version = "1.0.69", optional = true }
 tokio = { version = "1.53.1", default-features = false, optional = true }
-tracing = { version = "0.1", default-features = false, features = ["log", "std"] }
+tracing = { version = "0.1", default-features = false, features = [
+    "log",
+    "std",
+] }
 uriparse = { version = "0.6" }
-uuid-simd = { version = "0.8", default-features = false, features = ["std", "detect"] }
+uuid-simd = { version = "0.8", default-features = false, features = [
+    "std",
+    "detect",
+] }
 
 [build-dependencies]
 protobuf-codegen = { version = "3.7.2" }
@@ -160,6 +170,10 @@ required-features = ["up-l2-rpc-client", "up-l2-rpc-server", "util"]
 [[example]]
 name = "symphony_target"
 required-features = ["symphony", "up-l2-rpc-client", "up-l2-rpc-server", "util"]
+
+[[example]]
+name = "usubscription_server"
+required-features = ["usubscription", "up-l2-rpc-server", "util"]
 
 [[test]]
 name = "tck_uuri"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,19 +60,29 @@ util = ["tokio/sync"]
 [dependencies]
 async-trait = { version = "0.1.92" }
 bytes = { version = "1.12.1" }
+chrono = { version = "0.4.45" }
 mediatype = "0.23.0"
 mockall = { version = "0.15.0", optional = true }
 protobuf = { version = "3.7.2", features = ["with-bytes"], optional = true }
 rand = { version = "0.10.2" }
-regex = { version = "1.13.1", default-features = false, features = ["perf", "std"] }
+regex = { version = "1.13.1", default-features = false, features = [
+    "perf",
+    "std",
+] }
 serde = { version = "1.0.228", optional = true }
 serde_json = { version = "1.0.150", optional = true }
 symphony = { version = "0.1.5", optional = true }
 thiserror = { version = "2.0.20" }
 tokio = { version = "1.53.1", default-features = false, optional = true }
-tracing = { version = "0.1.44", default-features = false, features = ["log", "std"] }
+tracing = { version = "0.1.44", default-features = false, features = [
+    "log",
+    "std",
+] }
 uriparse = { version = "0.6.4" }
-uuid-simd = { version = "0.8.0", default-features = false, features = ["std", "detect"] }
+uuid-simd = { version = "0.8.0", default-features = false, features = [
+    "std",
+    "detect",
+] }
 
 [build-dependencies]
 protobuf-codegen = { version = "3.7.2" }
@@ -161,6 +171,10 @@ required-features = ["up-l2-rpc-client", "up-l2-rpc-server", "util"]
 [[example]]
 name = "symphony_target"
 required-features = ["symphony", "up-l2-rpc-client", "up-l2-rpc-server", "util"]
+
+[[example]]
+name = "usubscription_server"
+required-features = ["usubscription", "up-l2-rpc-server", "util"]
 
 [[test]]
 name = "tck_uuri"
--- a/examples/usubscription_server.rs
+++ b/examples/usubscription_server.rs
@@ -1,0 +1,144 @@
+/********************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+/*!
+This example illustrates how a service provider implements the server side of the
+uSubscription service.
+
+A [`RpcClientUSubscription`] drives the example by issuing `subscribe`/`unsubscribe`
+calls over an in-process [`LocalTransport`].
+ */
+
+use std::{collections::HashSet, sync::Arc};
+
+use tokio::sync::Mutex;
+use up_rust::{
+    communication::{
+        InMemoryRpcClient, InMemoryRpcServer, RequestHandler, RpcServer, ServiceInvocationError,
+        SubscriptionStatus, UPayload,
+    },
+    core::usubscription::{
+        extract_usubscription_request, pack_usubscription_response, RpcClientUSubscription,
+        SubscribeResponse, USubscription, USubscriptionRequest, USubscriptionResponse,
+        UnsubscribeResponse, RESOURCE_ID_SUBSCRIBE, RESOURCE_ID_UNSUBSCRIBE, USUBSCRIPTION_TYPE_ID,
+        USUBSCRIPTION_VERSION_MAJOR,
+    },
+    local_transport::LocalTransport,
+    StaticUriProvider, UAttributes, UUri,
+};
+
+/// A minimal, in-memory implementation of the uSubscription service.
+///
+/// It keeps track of the `(subscriber, topic)` pairs it has seen so far. A real
+/// implementation would persist this state and enforce the uSubscription state machine.
+#[derive(Default)]
+struct MyUSubscriptionService {
+    subscriptions: Mutex<HashSet<(String, String)>>,
+}
+
+#[async_trait::async_trait]
+impl RequestHandler for MyUSubscriptionService {
+    async fn handle_request(
+        &self,
+        resource_id: u16,
+        message_attributes: &UAttributes,
+        request_payload: Option<UPayload>,
+    ) -> Result<Option<UPayload>, ServiceInvocationError> {
+        // Decode the payload of *any* uSubscription operation into a typed request. Malformed or
+        // unsupported requests result in a `ServiceInvocationError`.
+        let request =
+            extract_usubscription_request(resource_id, message_attributes, request_payload)?;
+
+        let response: USubscriptionResponse = match request {
+            USubscriptionRequest::Subscribe(req) => {
+                println!(
+                    "SUBSCRIBE   subscriber={}, topic={}, expiration={:?}, sample_period={:?}",
+                    req.subscriber, req.topic, req.expiration, req.sample_period
+                );
+                self.subscriptions
+                    .lock()
+                    .await
+                    .insert((req.subscriber.to_string(), req.topic.to_string()));
+                SubscribeResponse {
+                    topic: req.topic,
+                    status: SubscriptionStatus::Subscribed,
+                }
+                .into()
+            }
+            USubscriptionRequest::Unsubscribe(req) => {
+                println!(
+                    "UNSUBSCRIBE subscriber={}, topic={}",
+                    req.subscriber, req.topic
+                );
+                self.subscriptions
+                    .lock()
+                    .await
+                    .remove(&(req.subscriber.to_string(), req.topic.to_string()));
+                UnsubscribeResponse.into()
+            }
+            // `USubscriptionRequest` is `#[non_exhaustive]`, so new operations can be
+            // added in future releases without breaking this code.
+            other => {
+                return Err(ServiceInvocationError::Unimplemented(format!(
+                    "operation not supported by this service: {other:?}"
+                )))
+            }
+        };
+
+        // Serialize the native response into the RPC response payload.
+        let payload = pack_usubscription_response(response)?;
+        Ok(Some(payload))
+    }
+}
+
+#[tokio::main]
+pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Using the LocalTransport lets us run the service and its client in the same
+    // process. A real deployment would use a networked transport (e.g. MQTT5 or Zenoh).
+    let transport = Arc::new(LocalTransport::default());
+
+    // The uSubscription service lives at a well-known address: empty (local) authority,
+    // the reserved entity ID and the service's major version.
+    let service_uri_provider = Arc::new(StaticUriProvider::new(
+        "",
+        USUBSCRIPTION_TYPE_ID,
+        USUBSCRIPTION_VERSION_MAJOR,
+    )?);
+
+    // Stand up the server and register our service for the subscribe/unsubscribe methods.
+    let rpc_server = InMemoryRpcServer::new(transport.clone(), service_uri_provider);
+    let service = Arc::new(MyUSubscriptionService::default());
+    rpc_server
+        .register_endpoint(None, RESOURCE_ID_SUBSCRIBE, service.clone())
+        .await?;
+    rpc_server
+        .register_endpoint(None, RESOURCE_ID_UNSUBSCRIBE, service.clone())
+        .await?;
+
+    // Now act as a client that subscribes to and unsubscribes from a topic. The
+    // subscriber's identity (its source address) is taken from this URI provider and
+    // ends up in `SubscribeRequest::subscriber` on the server side.
+    let client_uri_provider = Arc::new(StaticUriProvider::new("my-vehicle", 0xABCD, 0x01)?);
+    let rpc_client = Arc::new(InMemoryRpcClient::new(transport, client_uri_provider).await?);
+    let usubscription = RpcClientUSubscription::new(rpc_client);
+
+    let topic = UUri::try_from_parts("my-vehicle", 0x0000_800A, 0x01, 0x8001)?;
+
+    let status = usubscription.subscribe(&topic, None, None).await?;
+    println!("client: subscribe call returned status {status:?}");
+
+    usubscription.unsubscribe(&topic).await?;
+    println!("client: unsubscribe call completed");
+
+    Ok(())
+}
--- a/examples/usubscription_server.rs
+++ b/examples/usubscription_server.rs
@@ -1,0 +1,152 @@
+/********************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+/*!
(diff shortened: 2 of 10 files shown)


Cargo.toml
```diff
@@ -57,34 +57,40 @@ usubscription = ["up-l2-rpc-client"]
 util = ["tokio/sync"]
 
 [dependencies]
-async-trait = { version = "0.1" }
-bytes = { version = "1.11.1" }
-mediatype = "0.20"
-mockall = { version = "0.13", optional = true }
+async-trait = { version = "0.1.91" }
+bytes = { version = "1.12.1" }
```
sophokles73, line 61: for crates that have a major version > 0, I'd rather keep the lower minor version requirements in order to improve compatibility with older versions. For crates with a major version of 0 it doesn't make a difference because SemVer does not require minor versions to be backwards compatible in this case.

src/core/usubscription/usubscription_server.rs
```diff
@@ -0,0 +1,283 @@
...
+#[derive(Clone, Debug, PartialEq)]
+pub struct SubscribeRequest {
+    /// The uEntity that wants to subscribe (taken from the message's source address).
+    pub subscriber: UUri,
+    /// The topic to subscribe to.
+    pub topic: UUri,
+    /// The point in time at which the subscription expires (milliseconds since the Unix epoch).
+    pub expiration: Option<u64>,
+    /// The minimum duration (in milliseconds) between two events.
+    pub sample_period: Option<u32>,
```
sophokles73: how about using specific types here that actually represent the concept, e.g. using `chrono::DateTime` and `chrono::TimeDelta`? I guess that would make using the values less error prone and more intuitive ... and we could stop always explaining the unit of measure ...

src/core/usubscription/usubscription_server.rs
```diff
@@ -0,0 +1,283 @@
...
+/// Crate-internal glue that maps a protobuf request message to its public,
+/// protobuf-free counterpart.
+///
+/// Implementing this trait for a public request type is all that is needed to
+/// make it decodable via [`extract_request`]: the associated [`Proto`](Self::Proto)
+/// type selects which protobuf message to deserialize, and [`from_proto`](Self::from_proto)
+/// performs the mapping.
+trait FromProtoRequestPayload: Sized {
+    /// The protobuf message this request is transmitted as on the wire.
+    type Proto: MessageFull;
```
sophokles73: can we use `up_rust::ProtobufMappable` instead of `protobuf::MessageFul`? You should be able to achieve the same thing without adding another dependency on the `protobuf` crate ...

src/core/usubscription.rs
```diff
@@ -20,28 +20,34 @@ use crate::{communication::SubscriptionStatus, UStatus, UUri};
 #[cfg(all(feature = "up-l2-rpc-client", feature = "up-core-types"))]
 mod usubscription_client;
 #[cfg(all(feature = "up-l2-rpc-client", feature = "up-core-types"))]
+mod usubscription_server;
```
sophokles73: my understanding is that the types exported from this module are supposed to be used by implementors of the uSubscription service, right?
If so, then IMHO we should drop the up-l2-rpc-client feature requirement.

src/core/usubscription.rs
```diff
@@ -20,28 +20,34 @@ use crate::{communication::SubscriptionStatus, UStatus, UUri};
...
+#[cfg(all(feature = "up-l2-rpc-client", feature = "up-core-types"))]
 pub use usubscription_client::RpcClientUSubscription;
+#[cfg(all(feature = "up-l2-rpc-client", feature = "up-core-types"))]
+pub use usubscription_server::{
+    extract_usubscription_request, pack_usubscription_response, SubscribeRequest,
+    SubscribeResponse, USubscriptionRequest, USubscriptionResponse, UnsubscribeRequest,
+    UnsubscribeResponse,
+};
```
sophokles73: my understanding is that the types exported from this module are supposed to be used by implementors of the uSubscription service, right?
If so, then IMHO we should drop the up-l2-rpc-client feature requirement.

src/communication.rs
```diff
@@ -114,24 +114,30 @@ pub enum SubscriptionStatus {
...
+    impl From<&SubscriptionStatusProto> for SubscriptionStatus {
+        fn from(value: &SubscriptionStatusProto) -> Self {
+            match value {
+                UNSUBSCRIBED => SubscriptionStatus::Unsubscribed,
+                SUBSCRIBE_PENDING => SubscriptionStatus::SubscribePending,
+                SUBSCRIBED => SubscriptionStatus::Subscribed,
+                UNSUBSCRIBE_PENDING => SubscriptionStatus::UnsubscribePending,
+            }
+        }
+    }
```
sophokles73: You were able to change this from `TryFrom` to `From` because you no longer follow the recommended way for defining Enums in proto3, right? 
https://protobuf.dev/programming-guides/proto3/#enum

AnotherDaniel: Uh yes - but not intentionally. Wasn't aware of the guidance for enums, therefore didn't see a point in a value that is not part of the spec.
So, add it back in?

sophokles73: yes, I would suggest so

Cargo.toml
AnotherDaniel: So should we bump <0 major version dependencies?

src/core/usubscription/usubscription_server.rs
AnotherDaniel: Good idea, done.

src/core/usubscription/usubscription_server.rs
AnotherDaniel: Can do - but there are some other protobuf types used in this file, so this will just get rid of the MessageFull type.

src/core/usubscription.rs
AnotherDaniel: done

src/core/usubscription.rs
AnotherDaniel: done

Cargo.toml
sophokles73: Hmm, having thought about this a little more, I guess you should simply forget my previous comment. It doesn't make a difference whether we keep the lower minor version as a minimum requirement or state the most recent minor version.
If client code wants to use a more recent (minor) version, we depend on SemVer's backwards compatibility claim to make sure that up-rust will also work with the newer version. The same is true for the other direction, i.e. client code wants to use a version with an lower minor version number.

However, IMHO we should have the version updates in a separate commit on its own.

AnotherDaniel: ok - I've reverted the version bumps here

sophokles73:
@AnotherDaniel I have put most (if not all) of your proposed crate updates into a separate PR (https://github.com/eclipse-uprotocol/up-rust/pull/362). Can you take a look, approve and then rebase this PR?

sophokles73:
@AnotherDaniel can you also resolve the conflicts with main?

Cargo.toml
```diff
@@ -58,35 +58,40 @@ util = ["tokio/sync"]
...
-# addresses RUSTSEC-2026-0204
-crossbeam-epoch = { version = "0.9.20" }
 cucumber = { version = "0.23.0", features = ["output-junit"] }
 hex = { version = "0.4.3" }
-junit-report = { version = "0.9.1" }
-mockall = { version = "0.15.0" }
-test-case = { version = "3.3.1" }
-tokio = { version = "1.53.1", default-features = false, features = [
+mockall = { version = "0.13.1" }
+test-case = { version = "2.0.2" }
```
sophokles73: Any particular reason why you need/want to use an older version?

AnotherDaniel: I think you looked just as I was in the middle of battling git.

src/communication.rs
```diff
@@ -33,6 +33,7 @@ pub use simple_notifier::SimpleNotifier;
 mod pubsub;
 #[cfg(any(test, feature = "test-util"))]
 pub use pubsub::MockSubscriptionChangeHandler;
+#[cfg(feature = "up-l2-api")]
```
sophokles73: this is always the case because the whole _communication_ module is only defined if the `up-l2-api` is enabled

src/communication/pubsub.rs
```diff
@@ -17,10 +17,13 @@ use async_trait::async_trait;
 use thiserror::Error;
 
 use crate::{
-    communication::{CallOptions, RegistrationError, SubscriptionStatus, UPayload},
+    communication::{CallOptions, RegistrationError, UPayload},
     UListener, UStatus, UUri,
 };
 
+#[cfg(any(test, feature = "test-util", feature = "up-l2-api"))]
```
sophokles73: I guess you have added this in order to make sure that the MockSubscriptionChangeHandler also sees SubscriptionStatus. However, that is already the case, because the _pubsub_ module (and hence the MockSubscriptionChangeHandler) are only defined, if the `up-l2-api` feature is enabled.

src/core/usubscription.rs
```diff
@@ -21,37 +22,49 @@ use crate::{communication::SubscriptionStatus, UStatus, UUri};
 mod usubscription_client;
 #[cfg(all(feature = "up-l2-rpc-client", feature = "up-core-types"))]
 pub use usubscription_client::RpcClientUSubscription;
+#[cfg(feature = "up-core-types")]
```
sophokles73: this (and the `cfg`s on the other `mod` and `use` declarations) should not be necessary, because this module is only defined, if the `usubscription` feature is enabled, which includes `up-l2-rpc-client` -> `[up-l2-api, up-core-types, ...]` already

src/core/usubscription.rs
```diff
@@ -21,37 +22,49 @@ use crate::{communication::SubscriptionStatus, UStatus, UUri};
...
+pub use usubscription_server::{
+    extract_usubscription_request, pack_usubscription_response, USubscriptionRequest,
+    USubscriptionResponse,
+};
+#[cfg(feature = "up-core-types")]
+mod usubscription_proto;
 
 /// The uEntity (type) identifier of the uSubscription service.
 pub const USUBSCRIPTION_TYPE_ID: u32 = 0x0000_0000;
```
sophokles73: the type ID is actually defined as a u16 in the spec, in order to be super clear about that, I would therefore prefer to use a u16 here ...

src/core/usubscription.rs
```diff
@@ -136,15 +149,81 @@ impl SubscriptionInfo {
...
+    // pub subscriber: UUri,
+    /// The topic to unsubscribe from.
+    pub topic: UUri,
+}
+
+/// The response to an [`UnsubscribeRequest`].
+///
+/// The uSubscription service returns an empty message for this operation.
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct UnsubscribeResponse;
```
sophokles73: IMHO we do not need this, do we?

src/core/usubscription.rs
```diff
@@ -136,15 +149,81 @@ impl SubscriptionInfo {
...
+
+/// A request to reset uSubscription service internals.
+#[derive(Clone, Debug, PartialEq)]
+pub struct ResetRequest;
+
+/// The response to an [`ResetRequest`].
+///
+/// The uSubscription service returns an empty message for this operation.
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct ResetResponse;
```
sophokles73: same here, IMHO we do not need these

src/core/usubscription.rs
```diff
@@ -216,87 +288,41 @@ pub trait USubscription: Send + Sync {
     /// Returns an error if the attempt to unsubscribe has failed.
     async fn unsubscribe(&self, topic: &UUri) -> Result<(), UStatus>;
 
-    /// Gets all (currently) active subscriptions for a given topic.
+    /// Gets information for all current subscriptions; can be filtered according to optional of topic
+    /// and/or subscriber wildcard UURis.
```
sophokles73: IMHO this needs to be more specific as to how the filters are used. Do returned subscriptions need to match both (if specified), can I actually specify filters for both, or is it either or?

src/core/usubscription.rs
```diff
@@ -201,8 +273,8 @@ pub trait USubscription: Send + Sync {
     async fn subscribe(
         &self,
         topic: &UUri,
-        expiration: Option<u64>,
-        min_sample_period: Option<u32>,
+        expiration: Option<DateTime<Utc>>,
+        min_sample_period: Option<TimeDelta>,
```
sophokles73, line 248: IMHO we need to clarify in the Rust Doc that TimeDeltas will be clamped to [0; u32::MAX] milliseconds ...

src/core/usubscription/usubscription_proto.rs
```diff
@@ -0,0 +1,817 @@
...
+fn require_topic(topic: MessageField<UUriProto>) -> Result<UUri, UStatus> {
+    let topic = topic
+        .into_option()
+        .ok_or_else(|| UStatus::fail_with_code(UCode::InvalidArgument, "missing topic"))?;
+    UUri::try_from(&topic).map_err(|e| {
+        UStatus::fail_with_code(UCode::InvalidArgument, format!("invalid topic URI: {e}"))
+    })
+}
+
+/// Extracts and validates the topic from a protobuf request, failing if it is
```
sophokles73:
```suggestion
/// Extracts and validates the status from a protobuf request, failing if it is
```

src/core/usubscription/usubscription_client.rs
```diff
@@ -219,75 +50,34 @@ impl RpcClientUSubscription {
...
-                (None, None) => None.into(),
-                _ => Some(SubscribeAttributes {
-                    expire: unix_epoch_millis_as_protobuf_timestamp(expiration)?.into(),
-                    sample_period_ms: min_sample_period,
-                    ..Default::default()
-                })
-                .into(),
-            },
-            ..Default::default()
+        let subscription_request = SubscribeRequest {
```
sophokles73, line 64: FMPOV we should do some parameter validation here, e.g. check if expiration is in the past or sample_period is negative, is the topic really a topic?

examples/usubscription_server.rs
```diff
@@ -0,0 +1,146 @@
...
+    StaticUriProvider, UAttributes, UUri,
+};
+
+/// A minimal, in-memory implementation of the uSubscription service.
+///
+/// It keeps track of the `(subscriber, topic)` pairs it has seen so far. A real
+/// implementation would persist this state and enforce the uSubscription state machine.
+#[derive(Default)]
+struct MyUSubscriptionService {
+    subscriptions: Mutex<HashSet<(String, String)>>,
```
sophokles73: a RwLock would do here, I guess ;-)

src/communication.rs
AnotherDaniel: good catch - I'm a bit overwhelmed by the feature tree we have going :)

src/communication/pubsub.rs
AnotherDaniel: you're right - removed that line.

src/core/usubscription.rs
AnotherDaniel: I didn't originally add most of these, but are glad to remove them.

src/core/usubscription.rs
AnotherDaniel: ok

src/core/usubscription.rs
AnotherDaniel: You mean, use the unit type instead?

src/core/usubscription.rs
AnotherDaniel: Added some more description

src/core/usubscription.rs
AnotherDaniel: Added this info

src/core/usubscription/usubscription_proto.rs
AnotherDaniel: Done

examples/usubscription_server.rs
AnotherDaniel: done

src/core/usubscription/usubscription_client.rs
AnotherDaniel: ok

src/core/usubscription.rs
sophokles73: Yes, we only need the empty messages in the protobuf domain. We should not pollute our Rust traits with these types ...

src/core/usubscription.rs
sophokles73: But then you decided otherwise?

sophokles73:
It seems like you didn't push any of the changes that you mentioned in the comments ...

AnotherDaniel:
> It seems like you didn't push any of the changes that you mentioned in the comments ...

Yes, you're right. git mishap - should be back now.

src/core/usubscription.rs
```diff
@@ -185,24 +226,25 @@ pub trait USubscription: Send + Sync {
     /// # Parameters
     ///
     /// * `topic` - The topic to subscribe to.
-    /// * `expiration` - The point in time at which the subscription expires (milliseconds since Unix epoch).
+    /// * `expiration` - The point in time at which the subscription expires.
```
sophokles73, line 223: What if I supply a datetime in the past?

src/core/usubscription.rs
```diff
@@ -185,24 +226,25 @@ pub trait USubscription: Send + Sync {
...
     ///   If not specified, the subscription is valid until explicitly unsubscribed.
-    /// * `min_sample_period` - The minimum duration (in seconds) between two events that should be maintained
+    /// * `min_sample_period` - The minimum duration between two events that should be maintained
     ///   for remote only topics. Device dispatchers (i.e. streamers) use this attribute to reduce the
     ///   publication rates of events sent between devices.
     ///   This attribute is commonly used for mobile/cloud components subscribing to vehicle topics that are published
     ///   at a high rate. If the desired sampling period set by the subscriber is lower than the original publisher's
     ///   publication period, the attribute is ignored.
     ///   If not specified, the sampling period is set by the publisher.
+    ///   TimeDeltas used in `min_sample_period` will be clamped to [0; u32::MAX] milliseconds.
```
sophokles73, line 239: What is the meaning of a TimeDelta of 0? In fact, who cares about this value anyway? The streamer? The publishing uEntity will most likely not care at all, right?
Is there a (reasonable) minimal delta that we would like to impose, maybe something like 100ms? We are talking about remote topics only here, right?

src/core/usubscription/usubscription_client.rs
```diff
@@ -300,135 +93,57 @@ impl USubscription for RpcClientUSubscription {
...
-        subscriber: &UUri,
+        topic_filter: Option<UUri>,
+        subscriber_filter: Option<UUri>,
     ) -> Result<Vec<SubscriptionInfo>, UStatus> {
-        let subscriber_info = crate::up_core_api::usubscription::SubscriberInfo {
-            uri: Some(subscriber.into()).into(),
-            ..Default::default()
+        let fetch_subscriptions_request = FetchSubscriptionsRequest {
+            topic_filter: topic_filter.clone(),
+            subscriber_filter: subscriber_filter.clone(),
```
sophokles73: any reason why you need to clone these?

src/core/usubscription/usubscription_server.rs
```diff
@@ -0,0 +1,157 @@
...
+/// # Errors
+///
+/// Returns an error if `resource_id` does not identify a supported operation,
+/// or if the response object cannot be serialized into the expected type.
+pub fn pack_usubscription_response(
+    usubscription_response: USubscriptionResponse,
+) -> Result<Option<UPayload>, UStatus> {
+    match usubscription_response {
+        USubscriptionResponse::Subscribe(response) => Ok(Some(
+            UPayload::try_from_protobuf(SubscribeResponseProto::from(&response))
```
sophokles73:
```suggestion
            UPayload::try_from_protobuf(response)
```

should be sufficient here because SubscribeResponse is already ProtobufMappable ...

src/core/usubscription.rs
AnotherDaniel: Alright

src/core/usubscription.rs
AnotherDaniel: added more text from the spec

src/core/usubscription.rs
AnotherDaniel: Fair point, but I'd rather have the underlying discussion in up-spec context, before we address the language library.

src/core/usubscription/usubscription_client.rs
AnotherDaniel: Not anymore - good catch

src/core/usubscription/usubscription_server.rs
AnotherDaniel: done

src/core/usubscription/usubscription_proto.rs
```diff
@@ -161,6 +161,42 @@ impl TryFrom<&SubscriptionInfoProto> for SubscriptionInfo {
...
 }
 
+impl ProtobufMappable for SubscriptionInfo {
+    fn parse_from_packed_protobuf_bytes(proto: &[u8]) -> Result<Self, crate::SerializationError> {
+        SubscriptionInfo::try_from(&SubscriptionInfoProto::parse_from_bytes(proto)?)
+            .map_err(|e| SerializationError::new(e.to_string()))
+    }
+
+    fn parse_from_protobuf_bytes(proto: &[u8]) -> Result<Self, crate::SerializationError> {
+        Any::parse_from_bytes(proto)
```
sophokles73, line 171: _packed_ means _wrapped in Any_, so this looks more like the implementation of `parse_from_packed_protobuf_bytes`

It looks like you have confused these in all your ProtobufMappable implementations ...

src/core/usubscription/usubscription_proto.rs
```diff
@@ -161,6 +161,42 @@ impl TryFrom<&SubscriptionInfoProto> for SubscriptionInfo {
...
+    }
+
+    fn write_to_packed_protobuf_bytes(&self) -> Result<Vec<u8>, crate::SerializationError> {
+        Ok(SubscriptionInfoProto::try_from(self)
+            .map_err(|e| SerializationError::new(format!("failed to serialize to protobuf: {e}")))?
+            .write_to_bytes()?)
+    }
+
+    fn write_to_protobuf_bytes(&self) -> Result<Vec<u8>, crate::SerializationError> {
+        Any::pack(&SubscriptionInfoProto::try_from(self).map_err(|e| {
```
sophokles73, line 192: see above

src/core/usubscription/usubscription_client.rs
sophokles73: Did you add some validation logic? I cannot seem to find/see it ...

sophokles73 (changes requested):

src/core/usubscription/usubscription_proto.rs
AnotherDaniel: Actually, I was already wondering about this... but didn't get around to drawing the conclusion ;)

src/core/usubscription/usubscription_client.rs
AnotherDaniel: added now

sophokles73 (approved):
LGTM

sophokles73:
Can you rebase this, i.e. **not** merge the main branch into this PR ....

PR #369, opened 2026-08-06
Remove unneeded Default trait requirement from extract_protobuf

AnotherDaniel:
The Default trait requirement for the target type parameter of extract_protobuf functionality isn't actually needed (and is a hindrance for the usubscription v4 work), so removing it.


sophokles73 (approved):
LGTM

PR #372, opened 2026-08-14
Improve documentation and test coverage for UAttributesValidator

sophokles73:
Added test cases verifying that UAttributesValidator correctly handles wildcard UUri instances in all message types. This ensures that the validator enforces the expected constraints on UUris across different contexts.


AnotherDaniel (approved):
LGTM

PR #379, opened 2026-08-22
docs: add guided tour and align LocalTransport registration

PLeVasseur:
This PR adds a task-oriented rustdoc guide for Communication Layer roles, direct Transport Layer use, and transport implementations, with runnable role examples, compiled direct-transport examples, and explicit Cargo feature setup.

The guide uses the real `LocalTransport` in role examples, explains L1 push/pull delivery and implementation-specific send completion semantics, maps `UTransport`, L1, and protocol bindings, and links the published transport crates.

It also aligns `LocalTransport` listener registration with the existing L1 contract: invalid source/sink filter pairs now return `InvalidArgument`, and repeated registration of the same listener and filters succeeds idempotently without duplicate dispatch. No public API, dependency, manifest, workflow, generated-source, or specification changes are included.

### DIFF
--- a/src/guide/README.md
+++ b/src/guide/README.md
@@ -1,16 +1,12 @@
-# Guide: tutorials for every audience
+# Guide
 
-The pages, each self-contained, each with code:
-
-* [Writing an application](crate::guide::applications) — hub, with
-  [the Communication Layer](crate::guide::applications::communication)
-  (the default) and
+* [Writing an application](crate::guide::applications) — overview, with
+  runnable Communication Layer role examples and examples that use
   [the Transport Layer](crate::guide::applications::transport) directly.
 * [Implementing a transport](crate::guide::utransport) — the
-  [`UTransport`] contract from the provider side.
-* [The trait map](crate::guide::trait_map) — how the traits fit together
-  and which ones are yours.
+  relationship between [`UTransport`](crate::UTransport), L1 requirements, and
+  protocol bindings, with links to published transport implementations.
+* [The trait map](crate::guide::trait_map) — how the public traits fit together.
 
-Not sure where you fit? If you're *using* uProtocol, start with
-[applications](crate::guide::applications); if a trait name brought you
-here, start with [the trait map](crate::guide::trait_map).
+Start with [applications](crate::guide::applications) to use uProtocol, or the
+[trait map](crate::guide::trait_map) for an API overview.
--- a/src/guide/README.md
+++ b/src/guide/README.md
@@ -1,0 +1,16 @@
+# Guide: tutorials for every audience
+
+The pages, each self-contained, each with code:
+
+* [Writing an application](crate::guide::applications) — hub, with
+  [the Communication Layer](crate::guide::applications::communication)
+  (the default) and
+  [the Transport Layer](crate::guide::applications::transport) directly.
+* [Implementing a transport](crate::guide::utransport) — the
+  [`UTransport`] contract from the provider side.
+* [The trait map](crate::guide::trait_map) — how the traits fit together
+  and which ones are yours.
+
+Not sure where you fit? If you're *using* uProtocol, start with
+[applications](crate::guide::applications); if a trait name brought you
+here, start with [the trait map](crate::guide::trait_map).
--- a/src/guide/applications.md
+++ b/src/guide/applications.md
@@ -74,10 +74,11 @@ transport such as
 [`up-transport-mqtt5`](https://crates.io/crates/up-transport-mqtt5), or
 [`up-transport-vsomeip`](https://crates.io/crates/up-transport-vsomeip).
 
-Publishers, notifiers, subscribers, RPC clients, and RPC servers can share one
-[`Arc`](std::sync::Arc) clone. The role implementations are generic over
-[`UTransport`](crate::UTransport), so application code does not need to name
-the concrete transport type:
+Publishers, notifiers, subscribers, RPC clients, and RPC servers can share the
+same [`UTransport`](crate::UTransport) instance through
+[`Arc`](std::sync::Arc) clones. The role implementations are generic over
+`UTransport`, so application code does not need to name the concrete transport
+type:
 
 ```rust
 # use std::sync::Arc;
--- a/src/guide/applications.md
+++ b/src/guide/applications.md
@@ -1,83 +1,122 @@
-# Writing an application
+# Build an application
 
-Applications talk uProtocol through one of **two layers**, and knowing
-which one you're on is the whole orientation:
+The application-facing APIs are:
 
 * **[The Communication Layer](crate::guide::applications::communication)**
-  (up-L2, feature `communication`) — role traits: `Publisher`,
-  `Subscriber`, `Notifier`, `RpcClient`, `RpcServer`. **This is the
-  recommended starting point.** Start here unless you have a reason not to.
-* **[The Transport Layer](crate::guide::applications::transport)**
-  (up-L1, no features) — build [`UMessage`](crate::UMessage)s yourself
-  and hand them to a [`UTransport`](crate::UTransport). The floor the
-  roles stand on; yours directly when you need minimal dependencies,
-  custom filtering, or direct control over message construction.
-
-Thirty seconds to first message, on the communication layer:
+  provides five role traits across four messaging patterns:
+  [`Publisher`](crate::communication::Publisher),
+  [`Subscriber`](crate::communication::Subscriber),
+  [`Notifier`](crate::communication::Notifier), and
+  [`RpcClient`](crate::communication::RpcClient) with
+  [`RpcServer`](crate::communication::RpcServer).
+* **[The Transport Layer](crate::guide::applications::transport)** provides
+  direct message-level control and is the interface implemented by transport
+  providers.
 
 ```rust
 use std::sync::Arc;
+use up_rust::communication::{CallOptions, PubSubError, Publisher, SimplePublisher, UPayload};
 use up_rust::local_transport::LocalTransport;
 use up_rust::{StaticUriProvider, UPayloadFormat};
-use up_rust::communication::{CallOptions, Publisher, SimplePublisher, UPayload};
 
 #[tokio::main]
-async fn main() -> Result<(), Box<dyn std::error::Error>> {
+async fn main() -> Result<(), PubSubError> {
     let transport = Arc::new(LocalTransport::default());
-    let me = Arc::new(StaticUriProvider::new("my-vehicle", 0x1_0001, 1)?);
+    let identity = Arc::new(
+        StaticUriProvider::new("my-vehicle", 0x1_0001, 1)
+            .expect("valid publisher identity"),
+    );
 
-    SimplePublisher::new(transport, me)
-        .publish(0x8001, CallOptions::for_publish(None, None, None),
-                 Some(UPayload::new("92.5", UPayloadFormat::Text)))
+    SimplePublisher::new(transport, identity)
+        .publish(
+            0x8001,
+            CallOptions::for_publish(None, None, None),
+            Some(UPayload::new("92.5", UPayloadFormat::Text)),
+        )
         .await?;
-    // Delivered to every subscriber of my-vehicle/0x1_0001/1/0x8001.
     Ok(())
 }
 ```
 
-Everything below applies to both layers.
+## Choose an application role
 
-## Addresses in one box
+| Requirement | API | Additional dependency |
+| --- | --- | --- |
+| One-to-many events | [`Publisher`](crate::communication::Publisher) | — |
+| Receive published events | [`Subscriber`](crate::communication::Subscriber) | [`InMemorySubscriber`](crate::communication::InMemorySubscriber) uses uSubscription for bookkeeping and status |
+| One-way delivery to one uEntity | [`Notifier`](crate::communication::Notifier) | — |
+| Request/response | [`RpcClient`](crate::communication::RpcClient) and [`RpcServer`](crate::communication::RpcServer) | — |
+| Direct message send or receive | [`UTransport`](crate::UTransport) | — |
 
-Every example below uses [`UUri`](crate::UUri) literals; here is what
-they mean, once:
+## Address messages with UUri
+
+[`UUri`](crate::UUri) addresses identify an authority, uEntity, major version,
+and resource:
 
 ```text
 UUri::try_from_parts("my-vehicle", 0x1_0001, 1, 0x8001)
                       authority    entity   ver resource
 ```
 
-* **authority** — which device/domain the uEntity lives on.
-* **entity id** — which uEntity (a u32; deployments assign these).
-* **major version** — the uEntity's API major version.
-* **resource id** — *what* within the uEntity: `0x0000` is the uEntity
-  itself (the address RPC responses return to), ids `0x0001..=0x7FFF`
-  are RPC **methods**, and ids **`0x8000..=0xFFFE` are topics** a uEntity
-  publishes. `0xFFFF` is the wildcard. That is why the publish examples
-  use `0x8001` and the RPC endpoint uses `0x00A1`.
+* **authority** identifies the device or domain.
+* **entity id** identifies the uEntity.
+* **major version** identifies the uEntity API version.
+* **resource id** identifies a resource within the uEntity. `0x0000` addresses
+  the uEntity, `0x0001..=0x7FFF` identify RPC methods,
+  `0x8000..=0xFFFE` identify topics, and `0xFFFF` is the resource wildcard.
 
-## When things fail
+## Share a transport
 
-Every operation returns a code you can act on. Transport-level errors
-arrive as [`UStatus`](crate::UStatus) carrying a [`UCode`](crate::UCode):
-`Unavailable` (link down — retry with backoff), `AlreadyExists`
-(duplicate listener registration), `InvalidArgument` (a filter or
-message the transport cannot express), `PermissionDenied`. The roles
-wrap these: publishing fails with `PubSubError`, RPC with
-`ServiceInvocationError` — whose `CommStatus` variant means *the remote
-service* answered with a failure code, distinct from the transport
-failing to deliver. Timeouts surface as `CommStatus(DeadlineExceeded)`
-when the `ttl` you passed in `CallOptions` expires.
+The examples use [`LocalTransport`](crate::local_transport::LocalTransport), the
+crate's in-process push transport. A deployment can instead construct a
+transport such as
+[`up-transport-zenoh`](https://crates.io/crates/up-transport-zenoh),
+[`up-transport-mqtt5`](https://crates.io/crates/up-transport-mqtt5), or
+[`up-transport-vsomeip`](https://crates.io/crates/up-transport-vsomeip).
 
-## Where a real transport comes from
+Publishers, notifiers, subscribers, RPC clients, and RPC servers can share one
+[`Arc`](std::sync::Arc) clone. The role implementations are generic over
+[`UTransport`](crate::UTransport), so application code does not need to name
+the concrete transport type:
 
-The examples use `LocalTransport` so they can run anywhere. In a
-deployment, you construct one of the transport crates at startup —
-`up-transport-zenoh-rust`, `up-transport-mqtt5-rust`,
-`up-transport-vsomeip-rust`, and so on — configure it, wrap it in an
-`Arc`, and hand that same `Arc` to every role. Application code stays
-transport-agnostic: the role implementations are generic over
-`UTransport`, which is the point of the protocol.
+```rust
+# use std::sync::Arc;
+# use up_rust::local_transport::LocalTransport;
+let transport = Arc::new(LocalTransport::default());
+let publisher_transport = transport.clone();
+let rpc_transport = transport.clone();
+# let _ = (publisher_transport, rpc_transport);
+```
 
+See the [trait map](crate::guide::trait_map) for the public API relationships.
 
-For which trait is which, see the [trait map](crate::guide::trait_map).
+## Handle failures by API
+
+* [`UStatus`](crate::UStatus) is returned by Transport Layer operations. For
+  example, [`UTransport::send`](crate::UTransport::send) can reject an invalid
+  message with [`UCode::InvalidArgument`](crate::UCode::InvalidArgument) or
+  report an unavailable transport with
+  [`UCode::Unavailable`](crate::UCode::Unavailable).
+* [`RegistrationError`](crate::communication::RegistrationError) is returned
+  when a Communication Layer listener, subscriber, RPC client, or RPC endpoint
+  cannot be registered or unregistered.
+* [`PubSubError`](crate::communication::PubSubError),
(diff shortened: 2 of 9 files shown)


src/guide/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+The pages, each self-contained, each with code:
+
+* [Writing an application](crate::guide::applications) — hub, with
+  [the Communication Layer](crate::guide::applications::communication)
+  (the default) and
+  [the Transport Layer](crate::guide::applications::transport) directly.
+* [Implementing a transport](crate::guide::utransport) — the
+  [`UTransport`] contract from the provider side.
+* [The trait map](crate::guide::trait_map) — how the traits fit together
```
sophokles73: the links seem wrong to me, or am I missing something?

PLeVasseur: You are right that these do not resolve in the standalone GitHub Markdown rendering. They are intentional rustdoc intra-doc links: `src/guide/README.md` is included into the crate documentation from `src/lib.rs` using `#![doc = include_str!("guide/README.md")]`, and rustdoc resolves the `crate::...` paths.

I rebuilt the documentation with `cargo doc --no-deps --all-features` and checked the generated guide in a browser; the links navigate to the expected module pages. The generated documentation is working as intended, even though the paths look broken in the GitHub source view. I do not think a source change is needed unless we want this file to also function as a standalone GitHub README.

PLeVasseur: Yeah, these don't resolve in the GitHub Markdown rendering.

These are rustdoc intra-doc links. The way this works is that `src/guide/README.md` is included into the crate documentation from `src/lib.rs` by using `#![doc = include_str!("guide/README.md")]`, and then rustdoc resolves the `crate::...` paths.

Double-checked again and this seems to build and link correctly when I clicked around in it.

sophokles73: I guess this means that the guide only works as part of the generated Rust Docs but cannot be reasonably read from the filesystem or on GitHub, right?
Given that it is located under the `src` folder, that should not to be an issue, though ...

src/guide/utransport.md
```diff
@@ -0,0 +1,197 @@
+# The UTransport family
```
sophokles73: the example code here seems to basically be a reproduction of the LocalTransport that we have in the code base already. Is there a particular reason why we cannot simply refer to it here?

src/guide/applications.md
```diff
@@ -0,0 +1,83 @@
...
+service* answered with a failure code, distinct from the transport
+failing to deliver. Timeouts surface as `CommStatus(DeadlineExceeded)`
+when the `ttl` you passed in `CallOptions` expires.
+
+## Where a real transport comes from
+
+The examples use `LocalTransport` so they can run anywhere. In a
+deployment, you construct one of the transport crates at startup —
+`up-transport-zenoh-rust`, `up-transport-mqtt5-rust`,
+`up-transport-vsomeip-rust`, and so on — configure it, wrap it in an
```
sophokles73: how about linking to the repos or crates.io here?

src/guide/communication.md
```diff
@@ -0,0 +1,198 @@
+# The Communication Layer (up-L2)
+
+The roles hide message building entirely: you say *what* (publish this,
+call that) and the layer builds, addresses, and correlates the messages
+over whatever transport it was given. Enable `communication`.
```
sophokles73: not sure if all people correctly understand this to mean _enable the communication feature_ on the cargo command line ...

src/guide/utransport.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Payload and routing integrity
+
+Preserve all three payload states: absent, present with zero bytes, and present
+with nonempty bytes. Payload presence is not inferred from length.
+
+If the underlying protocol or broker mirrors routing fields in native headers,
+validate those fields against the message attributes before invoking a callback
+or forwarding. Route from the validated source and sink, never from payload
+bytes.
```
sophokles73: I am not sure if I understand the purpose of this paragraph. It does not really seem to provide any _actionable_ information to users, does it? It sounds more like a collection of bullet points that characterize a UTransport implementation's duties ...

src/guide/utransport.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Listener lifecycle and readiness
+
+Getting registration right matters more than it looks: a listener registered
+twice double-delivers, and one dropped early loses messages silently.
+Registration completes only when the binding can state what readiness means for
+its technology. Discovery-backed transports should expose a bounded peer or
+subscription readiness result, not use duplicate application sends as a
+readiness protocol. An absent peer returns a bounded, observable status.
```
sophokles73: I do not understand what this is supposed to mean

src/guide/utransport.md
```diff
@@ -0,0 +1,197 @@
...
+
+Listener dispatch preserves the binding's documented ordering and is bounded by
+an explicit queue, worker, or backpressure policy. Successful unregister is a
+quiescence boundary: a callback already running may complete if the binding says
+so, but no later callback may begin. Cancellation wakes blocking receive or poll
+operations, shutdown joins workers, and native entities are deleted
+deterministically so a transport can be recreated in the same domain.
```
sophokles73: This all sounds nice and reasonable. However, IMHO most of this cannot be traced back to the L1 spec. Maybe we should amend the spec accordingly?

sophokles73 (commented):
Thanks for adding this (long missing) guidance 👍 
There are only a few nits that need some consideration FMPOV.

src/guide/README.md
PLeVasseur: Yep, exactly :) The Markdown itself is still readable from the source, but the `crate::...` links only become working navigation once rustdoc generates the module pages.

Since these files live under `src` and are intended as rustdoc source rather than standalone GitHub documentation, I think that's okay too.

src/guide/utransport.md
PLeVasseur: Yeah, good point. The idea was to keep the guide self-contained and have a complete example that rustdoc could actually test. But maintaining a second transport implementation here gives us another place for the behavior to drift, even if the code continues to compile.

I've replaced it with a smaller runnable push-lifecycle example that uses the real `LocalTransport`, then walked through the actual implementation in prose with links to the relevant code instead of copying snippets. The guide now explicitly treats it as a push-mechanics example rather than an L1 conformance template, then explains the pull and/or push behavior a real broker, bus, or peer-to-peer binding adds. That keeps the walkthrough and compilation coverage without maintaining a second transport.

src/guide/applications.md
PLeVasseur: Yeah, good idea :) I've used the actual published crate names and linked them to crates.io. Those pages also link back to the repositories, so we get both without making this section too noisy.

src/guide/communication.md
PLeVasseur: Yeah, that's a good point :) I've made it explicit that these are Cargo features and added the command:

`cargo add up-rust --features communication,util`

`communication` enables the role APIs, and `util` is needed for the `LocalTransport` used in the examples.

src/guide/utransport.md
PLeVasseur: Yeah, I see what you mean. The idea was for this to be an implementation checklist, but it mostly says what a transport is responsible for without really helping someone implement it.

I've folded this into the walkthrough and tied each point to a concrete path: outbound validation and PDU mapping, pull `receive`, push dispatch, and listener registration/unregistration. The guide now points to the selected L1 protocol binding, keeps the L1 filter table authoritative over the SDK helper, and leaves transport-specific behavior in the transport's own documented contract.

src/guide/utransport.md
PLeVasseur: Yeah, that's fair. This was trying to squeeze a transport-specific discovery/readiness concern into one sentence, and it doesn't belong in the generic `UTransport` walkthrough. I've removed it.

src/guide/utransport.md
PLeVasseur: Yeah, I agree. I checked both the specification version pinned here and current up-spec `main`. L1 covers preserving message data, filter validation, idempotent registration, delivery after registration, invalid inbound messages, and stopping delivery after unregister. It doesn't currently define the broader ordering, backpressure, readiness, cancellation, or shutdown behavior described here.

I've trimmed the guide back to what we can trace to L1, distinguished L1's required error mappings from additional transport-specific failures that need documentation, and labeled the final checklist as minimum checks rather than complete conformance. It now covers the L1 push/pull distinction, listener limits, authorization controls, and priority mapping while pointing readers to the full pinned specification and selected binding. If we want the stronger lifecycle guarantees to apply consistently across transports, let's handle that as a separate up-spec follow-up first.

sophokles73 (approved):
LGTM

src/guide/README.md
PLeVasseur: Yep, exactly :) The Markdown itself is still readable from the source, but the `crate::...` links only become working navigation once rustdoc generates the module pages.

Since these files live under `src` and are intended as rustdoc source rather than standalone GitHub documentation, I think that's okay too.

src/guide/utransport.md
PLeVasseur: Yeah, good point. Maintaining a second transport implementation here gives us another place for the behavior to drift, even if the code continues to compile.

I've removed both the duplicate implementation and the prose tour of `LocalTransport`. The page now focuses on what a transport author needs from docs.rs: which delivery modes to implement, where the Rust/L1/protocol-binding contracts come from, what crosses the native technology boundary, and which completion semantics are easy to miss. It links the published Zenoh, MQTT 5, and SOME/IP transports as concrete design references instead of narrating another implementation.

src/guide/applications.md
PLeVasseur: Yeah, good idea :) I've used the actual published crate names and linked them to crates.io. Those pages also link back to the repositories, so we get both without making this section too noisy.

src/guide/communication.md
PLeVasseur: Yeah, that's a good point :) I've made it explicit that these are Cargo features and added the command:

`cargo add up-rust --features communication,util`

`communication` enables the role APIs, and `util` is needed for the `LocalTransport` used in the examples.

src/guide/utransport.md
PLeVasseur: Yeah, I see what you mean. The original checklist mostly said what a transport is responsible for without helping someone find the authority or make an implementation decision.

I've replaced it with a focused technology-boundary section. It identifies the Rust API, L1, and selected binding as separate authorities; explains outbound and inbound mapping and validation; and anchors push registration in observable filter and delivery behavior without prescribing one native subscription architecture. The page keeps the L1 filter table authoritative over the SDK helper and points to existing transport crates for concrete designs.

src/guide/utransport.md
PLeVasseur: Yeah, that's fair. This was trying to squeeze a transport-specific discovery/readiness concern into one sentence, and it doesn't belong in generic `UTransport` guidance. I've removed it.

src/guide/utransport.md
PLeVasseur: Yeah, I agree. I checked both the specification version pinned here and current up-spec `main`. L1 covers preserving message data, filter validation, idempotent registration, delivery after registration, invalid inbound messages, and stopping delivery after unregister. It doesn't currently define the broader ordering, backpressure, readiness, cancellation, or shutdown behavior described here.

I've removed the broad lifecycle guidance and the prose conformance checklist. The page now distinguishes the Rust trait, the complete pinned L1 specification, and the selected protocol binding as separate authorities, while retaining only the non-obvious completion semantics directly relevant to implementing the methods. Transport-specific reconnect, ordering, backpressure, and shutdown behavior is left to each transport's contract. If we want stronger common guarantees, let's handle that as a separate up-spec follow-up first.

src/guide/README.md
PLeVasseur: Yep, exactly :) The Markdown itself is still readable from the source, but the `crate::...` links only become working navigation once rustdoc generates the module pages.

Since these files live under `src` and are intended as rustdoc source rather than standalone GitHub documentation, I think that's okay too.

src/guide/utransport.md
PLeVasseur: Yeah, good point. Maintaining a second transport implementation here gives us another place for the behavior to drift, even if the code continues to compile.

I've removed both the duplicate implementation and the prose tour of `LocalTransport`. The page now focuses on what a transport author needs from docs.rs: which delivery modes to implement, where the Rust/L1/protocol-binding contracts come from, what crosses the native technology boundary, and which completion semantics are easy to miss. It links the published Zenoh, MQTT 5, and SOME/IP transports as concrete design references instead of narrating another implementation.

src/guide/applications.md
PLeVasseur: Yeah, good idea :) I've used the actual published crate names and linked them to crates.io. Those pages also link back to the repositories, so we get both without making this section too noisy.

src/guide/communication.md
PLeVasseur: Yeah, that's a good point :) I've made it explicit that these are Cargo features and added the command:

`cargo add up-rust --features communication,util`

`communication` enables the role APIs, and `util` is needed for the `LocalTransport` used in the examples.

src/guide/utransport.md
PLeVasseur: Yeah, I see what you mean. The original checklist mostly said what a transport is responsible for without helping someone find the authority or make an implementation decision.

I've replaced it with a focused technology-boundary section. It identifies the Rust API, L1, and selected binding as separate authorities; explains outbound and inbound mapping and validation; and anchors push registration in observable filter and delivery behavior without prescribing one native subscription architecture. The page keeps the L1 filter table authoritative over the SDK helper and points to existing transport crates for concrete designs.

src/guide/utransport.md
PLeVasseur: Yeah, that's fair. This was trying to squeeze a transport-specific discovery/readiness concern into one sentence, and it doesn't belong in generic `UTransport` guidance. I've removed it.

src/guide/utransport.md
PLeVasseur: Yeah, I agree. I checked both the specification version pinned here and current up-spec `main`. L1 covers preserving message data, filter validation, idempotent registration, delivery after registration, invalid inbound messages, and stopping delivery after unregister. It doesn't currently define the broader ordering, backpressure, readiness, cancellation, or shutdown behavior described here.

I've removed the broad lifecycle guidance and the prose conformance checklist. The page now distinguishes the Rust trait, the complete pinned L1 specification, and the selected protocol binding as separate authorities, while retaining only the non-obvious completion semantics directly relevant to implementing the methods. Transport-specific reconnect, ordering, backpressure, and shutdown behavior is left to each transport's contract. If we want stronger common guarantees, let's handle that as a separate up-spec follow-up first.

src/guide/applications.md
```diff
@@ -0,0 +1,122 @@
...
+[`UUri`](crate::UUri) addresses identify an authority, uEntity, major version,
+and resource:
+
+```text
+UUri::try_from_parts("my-vehicle", 0x1_0001, 1, 0x8001)
+                      authority    entity   ver resource
+```
+
+* **authority** identifies the device or domain.
+* **entity id** identifies the uEntity.
```
sophokles73, line 62: I always wonder whether we should not better split up the ue_id into two u16 fields (ue_type_d and ue_instance_id) in the model. IMHO the only reason why we have them both in one field is that protobuf does not have a u16 type ...
WDYT? I guess now would still be time to make that change in up-rust ...

src/guide/applications.md
```diff
@@ -0,0 +1,122 @@
...
+crate's in-process push transport. A deployment can instead construct a
+transport such as
+[`up-transport-zenoh`](https://crates.io/crates/up-transport-zenoh),
+[`up-transport-mqtt5`](https://crates.io/crates/up-transport-mqtt5), or
+[`up-transport-vsomeip`](https://crates.io/crates/up-transport-vsomeip).
+
+Publishers, notifiers, subscribers, RPC clients, and RPC servers can share one
+[`Arc`](std::sync::Arc) clone. The role implementations are generic over
+[`UTransport`](crate::UTransport), so application code does not need to name
+the concrete transport type:
```
sophokles73:
```suggestion
Publishers, notifiers, subscribers, RPC clients, and RPC servers can share the same [`UTransport`](crate::UTransport) instance wrapped in [`Arc`](std::sync::Arc) clones. The role implementations are generic over [`UTransport`](crate::UTransport), so application code does not need to name
the concrete transport type:
```

src/guide/utransport.md
```diff
@@ -0,0 +1,187 @@
...
+
+The binding defines the native representation. L1 defines what callers can
+observe at the `UTransport` methods.
+
+## Implement outbound send
+
+[`UTransport::send`](crate::UTransport::send) takes one complete `UMessage`.
+The implementation should:
+
+1. validate the message attributes;
```
sophokles73: I am actively working on making this requirement obsolete. UMessageBuilder only produces valid messages already, the remaining thing is the deserialization from protobuf and mapping from custom types (codecs). I am currently preparing a PR for addressing the latter ...

src/local_transport.rs
```diff
@@ -81,21 +83,15 @@ impl UTransport for LocalTransport {
         sink_filter: Option<&UUri>,
         listener: Arc<dyn UListener>,
     ) -> Result<(), UStatus> {
+        verify_filter_criteria(source_filter, sink_filter).map_err(|status| *status)?;
```
sophokles73, line 86: nice addition 👍

src/local_transport.rs
```diff
@@ -81,21 +83,15 @@ impl UTransport for LocalTransport {
...
-        if listeners.contains(&registered_listener) {
-            Err(UStatus::fail_with_code(
-                crate::UCode::AlreadyExists,
-                "listener already registered for filters",
-            ))
-        } else {
-            listeners.insert(registered_listener);
-            Ok(())
-        }
+        listeners.insert(registered_listener);
```
sophokles73, line 93: good change

sophokles73 (approved):
LGTM

src/guide/README.md
PLeVasseur: Yep, exactly :) The Markdown itself is still readable from the source, but the `crate::...` links only become working navigation once rustdoc generates the module pages.

Since these files live under `src` and are intended as rustdoc source rather than standalone GitHub documentation, I think that's okay too.

src/guide/utransport.md
PLeVasseur: Yeah, good point. Maintaining a second transport implementation here gives us another place for the behavior to drift, even if the code continues to compile.

I've removed both the duplicate implementation and the prose tour of `LocalTransport`. The page now links `LocalTransport` as a minimal in-process push implementation, then points users to the published Zenoh, MQTT 5, and SOME/IP transports. The rest of the page focuses on the delivery modes, Rust/L1/protocol-binding authorities, native technology boundary, and completion semantics a transport author needs.

src/guide/applications.md
PLeVasseur: Yeah, good idea :) I've used the actual published crate names and linked them to crates.io. Those pages also link back to the repositories, so we get both without making this section too noisy.

src/guide/communication.md
PLeVasseur: Yeah, that's a good point. I've made it explicit that these are Cargo features and added the command:

`cargo add up-rust --features communication,util`

`communication` enables the role APIs, and `util` is needed for the `LocalTransport` used in the examples.

src/guide/utransport.md
PLeVasseur: Yeah, I see what you mean. The original checklist mostly said what a transport is responsible for without helping someone find the authority or make an implementation decision.

I've replaced it with a guide explaining relevant boundaries. It identifies the Rust API, L1, and selected binding as separate authorities; explains outbound encoding, inbound decoding, and validation at `UMessage` construction boundaries; and gives method-specific push, pull, and error behavior without prescribing one native subscription architecture.

I have It also point to `LocalTransport` and the published transport crates for concrete designs.

src/guide/utransport.md
PLeVasseur: This was trying to squeeze a transport-specific discovery/readiness concern into one sentence, and it doesn't belong in generic `UTransport` guidance. I've removed it to clean this up.

src/guide/utransport.md
PLeVasseur: Yeah, totally agree. I've shifted the broader lifecycle concepts out of here.

I think I should go open an issue on up-spec to track adding these there.

src/guide/applications.md
PLeVasseur: Yeah, we're kind of still eating this artificial combination at the behest of a ProtoBuf, which we're trying to uncouple from as being assumed / needed.

I think it's worthwhile opening an issue on up-spec to flesh that out. Either you or I can sketch something in said issue to start with.

src/guide/applications.md
PLeVasseur: Good call. Updated in `a053303`: the text now says that all roles share the same `UTransport` instance through `Arc` clones, and the example shows those clones coming from the same `Arc`.

src/guide/utransport.md
PLeVasseur: Thanks for the heads-up. Updated in `a053303`: I removed the explicit validation step and validator example from `send`. The guide now puts validation at the boundary that constructs a `UMessage`: `UMessageBuilder` validates before returning one, and decoders or custom mappings should reject invalid representations there. The flow diagram likewise starts the outbound path with a validated `UMessage` and validates inbound data before producing one.

The pinned L1 version still specifies `InvalidArgument` if an invalid message reaches `send`, so the guide retains that only as a fallback for integrations that do not yet enforce the construction invariant, not as a reason for transports to duplicate validation.

PLeVasseur (commented):
Thanks for the thorough review, @sophokles73.

I've taken one more pass to clean things up.

Could you please take another look to make sure I've addressed what you brought up?

src/guide/applications.md
sophokles73: I have created https://github.com/eclipse-uprotocol/up-spec/issues/342

sophokles73 (commented):
LGTM

src/guide/applications.md
PLeVasseur: Sounds good. Thanks!

src/guide/transport.md
```diff
@@ -0,0 +1,130 @@
...
+
+assert!(any_source.matches(&one_topic));
+assert!(topic_at_any_source.matches(&one_topic));
+```
+
+Use an exact [`UUri`](crate::UUri) to select one address. Wildcard filters can
+select several sources or resources, subject to the valid source/sink resource
+combinations in the L1 Transport Layer specification.
+
+The [transport implementation guide](crate::guide::utransport) points to the
```
AnotherDaniel, line 126:
```suggestion
The [uTransport implementation guide](crate::guide::utransport) points to the
```

AnotherDaniel: To distinguish from the currently open transport guide document.

PR #380, opened 2026-08-24
Fix zizmor findings in CI/CD workflows

sophokles73:
Fixed all findings from the Zizmor security scan in the CI/CD workflows.


AnotherDaniel (approved):
ok

PR #381, opened 2026-08-25
Move expiration check to UUID

sophokles73:
The logic for checking if a message has expired has been moved to the UUID struct. This allows for a more centralized and consistent way of checking expiration across different parts of the codebase, in particular out of the context of existing UAttributes and UMessage structs.


AnotherDaniel (approved):
Makes sense

PR #383, opened 2026-08-27
Add validation step to UMessage constructor

sophokles73:
Added a validation step to the crate internal UMessage constructor. This ensures that any UMessage created through the constructor will be validated for correctness.


AnotherDaniel (approved):
LGTM

PR #389, opened 2026-09-17
Remove runtime dependency on regex

### LINKED ISSUE #388: Consider replacing the runtime dependency on regex with custom pattern matching function
We currently use the regex crate to validate UUri authority names by means of matching to a regex pattern.
However, the pattern is pretty straight forward so we might want to replace it with custom pattern matching logic based on standard Rust string functionality.

sophokles73:
Replaced usage of the regex crate with a custom function that performs the same pattern matching using standard string functionality.

Addresses #388


AnotherDaniel (approved):
Very worthwhile change

PR #391, opened 2026-09-17
Improve dependency check workflow

sophokles73:
This PR makes sure that the list of crates that we depend on is sorted, contains no duplicate entries and does not contain the up-rust crate itself.


AnotherDaniel (approved):
Probably ok - is there some description what this changes, even just a half-sentence for the PR maybe?

Issue #387, opened 2026-09-16
Consider replacing uriparse with fluent-uri crate


sophokles73:
The uriparse crate does not seem to have received any updates for the last four years. Maybe we should replacce it with https://crates.io/crates/fluent-uri which seems to be more actively maintained.

sophokles73:
closed via #387

PR #392, opened 2026-09-18
Replace uriparse crate with fluent-uri

sophokles73:
The uriparse crate has not seen any updates for the last 7 years, so we decided to replace it with the fluent-uri crate, which is actively maintained and provides a more modern API for URI handling.

Addresses #387


AnotherDaniel (approved):
LGTM

